Issue #160, opened 2014-11-21
Set option on Annotation to invalidate with BAN or PURGE


lepiaf:
Hi,

When I use @InvalidatePath, It always use PURGE request. I would like to have a choice to do BAN or PURGE.

I suggest to add an option "type" : 

``` php
/**
 * @InvalidateRoute(name="some_route", params={"id" = {"expression"="id"}}, type="ban")
 */

/**
 * @InvalidateRoute(name="some_route", params={"id" = {"expression"="id"}}, type="purge")
 */

/**
 * @InvalidatePath(path="/some/path/.*", type="ban")
 */

/**
 * @InvalidatePath(path="/some/path", type="purge")
 */
```

Then, in InvalidateSubscriber it will switch between `$this->cacheManager->invalidatePath()` and `$this->cacheManager->invalidateRegex()`

What do you think ?

dbu:
a simpler case could be to ban everything up to the first parameter. so for /products/{id} ban /products/ then we would only look at the pattern

PR #161, opened 2014-11-21
[WIP] Add type on annotation and handle it

lepiaf:
PR for https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/160

I have to replace UrlGenerator with Router because I need to know if a route has parameters, and then replace all token with a wildcard.

I also change method scope to protected. It make possible to fit for special need.


dbu:
ping

EventListener/InvalidationSubscriber.php
```diff
@@ -220,11 +226,60 @@ private function invalidateRoutes(array $routes, Request $request)
...
+     * Set all route params as widlcard by default
+     *
+     * @param string $routeName
+     *
+     * @return array
+     */
+    protected function preFillParams($routeName)
+    {
+        $params = array();
+        $route = $this->router->getRouteCollection()->get($routeName->getName());
```
stof: please don't do this. Using ``->getRouteCollection()`` at runtime kills your performance, because it triggers the loading of routes each time

EventListener/InvalidationSubscriber.php
```diff
@@ -220,11 +226,60 @@ private function invalidateRoutes(array $routes, Request $request)
...
+
+    /**
+     * Generate url regex by route path
+     *
+     * @param string $routeName
+     * @param array $params
+     *
+     * @return string
+     */
+    protected function generateRegexPath($routeName, $params)
```
stof: you should use private visibility, not protected one (protected means it is an extension point covered by BC concerns, which must be avoided whenever we can to simplify the maintenance).

And array arguments should be typehinted

lepiaf:
Hi, I don't have time to work on it now. Can you open an issue and let somone else to work on it?

EventListener/InvalidationSubscriber.php
```diff
@@ -220,6 +226,26 @@ private function invalidateRoutes(array $routes, Request $request)
                 }
             }
 
+            if (null === $route->getParams()) {
+                $route = $this->router->getRouteCollection()->get($route->getName());
```
stof: your code snippets don't do anything to prove @ddeboer is wrong (and this is logical, because he is not)

dbu:
we can just leave this open for now and see if somebody wants to work on it.

dbu:
@ddeboer what do we do with this one? i would tend to drop it, it seems very complicated. if there is somebody needing it, an option could be to add an annotation to ban on the static prefix of a route so there would be no need to integrate that deeply with the routing.

ddeboer:
Closing this for now. @lepiaf If you feel this is still necessary and have the time to work on it, please re-open. :smile:

Issue #181, opened 2014-12-13
tagging response while building the content


dbu:
We are looking into integrating this bundle with the cmf. One idea we had was to offer a twig function to register a tag from the templates (as cms can have complex structures and templates is the place we know what is really rendered). is that insane or a good idea? should we even offer something in here maybe?

A problem with that is that [tagResponse](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/CacheManager.php#L54) requires to pass the Response object. when rendering in twig, that does not exist yet. Could we offer another method on CacheManager that adds a tag to a response listener and adds the tags at that point?

see https://github.com/symfony-cmf/symfony-cmf/issues/211 for other discussion

TODO
* [x] allow to set tags without having a response object => #182
* [ ] twig function for adding tags => #211

See #184 for the rest of the cmf related discussion.

ddeboer:
#182 is done, so only the Twig extension remains. Should it go in the bundle or the library (or I guess ideally a separate twig-bridge package)? People may want to use the extension in a non-Symfony application.

dbu:
i would put it into the bundle. its just a shortcut to call a php
function in the proper twig way. if somebody needs it outside the
symfony full stack context, he can either depend on the bundle
nontheless and bootstrap the extension himself, or simply copy the 10
lines of code, i would say.

ddeboer:
Fair enough!

PR #182, opened 2014-12-13
Have a method to add tags to the TagSubscriber directly

dbu:
This is useful to tag responsens when the response object does not yet exist or is not known.

This solves part of #181

@ddeboer do you want me to add a twig extension to tag responses from inside twig templates here? or should we provide that from SymfonyCmfHttpCacheBundle rather? the cmf bundle will extend it to handle doctrine objects in addition to just strings.

TODO
* [x] Refactor TagSubscriber to use TagHandler
* [x] Adjust documentation to talk about TagHandler
* [x] Squash commits


ddeboer:
All right, so let’s try to group tags by the response they should be applied to.

dbu:
yeah, i wanted to look into that, but realized it would make a lot more sense to do that in a TagHandler than adding more non-event things into the Tag event listener class and started with https://github.com/FriendsOfSymfony/FOSHttpCache/pull/162 as precondition.

ddeboer:
I guess this can be finished now, since https://github.com/FriendsOfSymfony/FOSHttpCache/pull/162 and https://github.com/FriendsOfSymfony/FOSHttpCache/pull/171 have been merged.

Resources/doc/features/tagging.rst
```diff
@@ -35,15 +35,22 @@ Setting and Invalidating Tags
 You can tag responses in three ways: with the cache manager, configuration and
 annotations.
 
-Cache Manager
-~~~~~~~~~~~~~
+Tagging from code
+~~~~~~~~~~~~~~~~~
 
-Use ``tagResponse($response, $tags)`` to set tags on a response::
+Inject the ``CacheManager`` (service ``fos_http_cache.cache_manager``) and
```
dbu: TODO: adjust the doc to the new architecture

EventListener/TagSubscriber.php
```diff
@@ -37,6 +37,13 @@ class TagSubscriber extends AbstractRuleSubscriber implements EventSubscriberInt
     private $expressionLanguage;
 
     /**
+     * List of tags to add to response.
+     *
+     * @var string[]
+     */
+    private $tags = array();
```
dbu: TODO: refactor to use the TagHandler that was added to the library

EventListener/TagSubscriber.php
```diff
@@ -63,12 +86,16 @@ public function onKernelResponse(FilterResponseEvent $event)
     {
         $request = $event->getRequest();
         $response = $event->getResponse();
-
-        $tags = array();
+        if ($request->isMethodSafe()) {
```
dbu: @ddeboer what do we want to do about this? allow to invalidate tags by adding them to the tagHandler in an unsafe request, or throw an exception in this case and tell that to invalidate the user should explicitly call invalidate?

EventListener/TagSubscriber.php
```diff
@@ -86,8 +113,13 @@ public function onKernelResponse(FilterResponseEvent $event)
...
         if ($request->isMethodSafe()) {
-            // For safe requests (GET and HEAD), set cache tags on response
-            $this->cacheManager->tagResponse($response, $tags);
+            if ($event->isMasterRequest()) {
+                // For safe requests (GET and HEAD), set cache tags on response
+                $this->cacheManager->tagResponse($response, $tags);
+            } else {
+                // Retain tags to attach to response to master request
+                $this->tags = $tags;
+            }
```
dbu: i ended up doing this logic. otherwise tags from subrequests get lost. is this sane?

for ESI requests, everything is a separate master request and its own cache lifecycle, so having different tags makes sense.

dbu:
so i propose that we only tag the response to the master request as we otherwise lose the tags of subrequests. note that this will be a behaviour change for existing features: the tag annotation in the subrequest gets lost in the 1.2 bundle. i think the new behaviour is correct and consider the other thing a bug.

EventListener/TagSubscriber.php
ddeboer: Are you sure about ESI requests being master requests? I thought they were sub requests.

Resources/doc/features/tagging.rst
```diff
@@ -53,7 +60,28 @@ Use ``tagResponse($response, $tags)`` to set tags on a response::
         }
     }
 
-Then call ``invalidateTags($tags)`` to invalidate the tags you set::
+Or inject the ``TagSubscriber`` (service ``fos_http_cache.event_listener.tag``)
+and call ``addTags($tags)`` to set tags that will be added to the response once
+it exists::
```
ddeboer: “Or *if you have no access to a Response object* inject…”

ddeboer:
> so i propose that we only tag the response to the master request 

:+1:

EventListener/TagSubscriber.php
ddeboer: I prefer the exception, for it gets confusing for users if we let `addTags()` do invalidation too. The distinction makes sense: you have only one response during the request/response cycle that you want to set exactly the right set of tags on. While invalidating, however, you can just call `invalidateTags()` with some tag(s), and later in the cycle call it again with some other tag(s).

EventListener/TagSubscriber.php
dbu: pretty much so, yes: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L492 - this is also the code path reached when esi is done (its only considered a sub request inside the HttpCache but then passed as a new master request to HttpKernel - which seems to me the expected and correct behaviour)

dbu: @stof does this make sense to you or am i missing something? it seems to me we only want to tag on the response to master request.

the current implementation introduces an inconsistency in that tags from configuration / annotation are preserved from subrequest controllers, while cache control of subrequests is lost. but i think that is ok, as such information set on a response object is lost in twig subrequests in general.

stof: ESI requests are indeed master requests (they are subrequests inside Varnish/HttpCache, but as far as AppKernel is concerned, it is an incoming request)

stof: and when using ESI, each master request has its own cache, so you will want to tag the ESI requests on themselves

dbu: thanks @stof for confirming. my other question was whether you agree that we tag only on response to master request, not response to subrequests (you talked about using the request stack and associate tags with requests). it seems to me to make no sense to ever tag a response to a subrequest, as the headers of those responses are simply discarded.

EventListener/TagSubscriber.php
```diff
@@ -86,8 +113,13 @@ public function onKernelResponse(FilterResponseEvent $event)
         }
 
         if ($request->isMethodSafe()) {
-            // For safe requests (GET and HEAD), set cache tags on response
-            $this->cacheManager->tagResponse($response, $tags);
+            if ($event->isMasterRequest()) {
```
dbu: argh, this is only available in symfony 2.4+, but not 2.3. i will fix that line to work with 2.3

dbu:
@stof do you agree that we tag only on response to master request, not response to subrequests (you talked about using the request stack and associate tags with requests). it seems to me to make no sense to ever tag a response to a subrequest, as the headers of those responses are simply discarded.

CacheManager.php
```diff
@@ -67,6 +67,8 @@ public function setGenerateUrlType($generateUrlType)
      *                           response
      *
      * @return $this
+     *
+     * @deprecated Add tags with TagHandler::addTags and then use TagHandler::tagResponse
```
dbu: i don't go to the TagHandler here. so people who mix up new and old things might get into troubles. but otherwise we complicate things quite a bit.

Command/InvalidateTagCommand.php
```diff
@@ -32,13 +38,14 @@ class InvalidateTagCommand extends BaseInvalidateCommand
      * If no cache manager is specified explicitly, fos_http_cache.cache_manager
      * is automatically loaded.
      *
-     * @param CacheManager|null $cacheManager The cache manager to talk to.
-     * @param string            $commandName  Name of this command, in case you want to reuse it.
+     * @param TagHandler|null $tagHandler  The tag handlerto talk to.
+     * @param string          $commandName Name of this command, in case you want to reuse it.
      */
-    public function __construct(CacheManager $cacheManager = null, $commandName = 'fos:httpcache:invalidate:tag')
+    public function __construct(TagHandler $tagHandler = null, $commandName = 'fos:httpcache:invalidate:tag')
```
dbu: if people defined their own service for the InvalidateTagCommand or extended the command, this is a BC break. do we care about it or can we assume this an edge case? if we care, what are alternatives that make sense?

dbu: we can keep the BC method of CacheManager of course, but people using the new TagHandler then will be just as confused.

EventListener/TagSubscriber.php
```diff
@@ -73,24 +73,21 @@ public function onKernelResponse(FilterResponseEvent $event)
...
-        } else {
+            $this->tagHandler->addTags($tags);
+            if (HttpKernelInterface::MASTER_REQUEST === $event->getRequestType()) {
+                // For safe requests (GET and HEAD), set cache tags on response
+                $this->tagHandler->tagResponse($response);
+            }
+        } elseif (count($tags)) {
             // For non-safe methods, invalidate the tags
-            $this->cacheManager->invalidateTags($tags);
+            $this->tagHandler->invalidateTags($tags);
```
dbu: this is a bit of a weird interaction with the tag handler:

* for tagging the response, the only way to go is to do $tagHandler->addTags($tags); $tagHandler->tagResponse($response);
* for invalidating, the only way to go is to $tagHandler->invalidateTags($tags);

so tagging a response is only possible from the values in the handler, invalidating is only possible with outside tags.

also, as we have no way to reset the tags on tagResponse, we will keep the tags in the handler and if there is ever more than one response sent, tags will accumulate. should we add a reset method to the tagHandler and trigger that reset in tagResponse?

Handler/TagHandler.php
```diff
@@ -0,0 +1,34 @@
...
+    public function tagResponse(Response $response, $replace = false)
+    {
+        if (!$replace && $response->headers->has($this->getTagsHeaderName())) {
+            $header = $response->headers->get($this->getTagsHeaderName());
+            if ('' !== $header) {
+                $this->addTags(explode(',', $response->headers->get($this->getTagsHeaderName())));
+            }
+        }
+
+        if ($this->getTagsHeaderValue()) {
```
dbu: this is the only way to determine whether we actually have any tags in the handler. does that make sense, or should we add a count() method? or even a getTags()?

dbu:
@ddeboer had another go on this. i think things are shaping up, i added some comments on the code. glad for some feedback / opinion so that we can wrap this up.

dbu:
@ddeboer ping

Command/InvalidateTagCommand.php
ddeboer: The BC way would be to allow both CacheManager and TagHandler as constructor arguments, check the type and throw an exception if it is neither of both.

Handler/TagHandler.php
ddeboer: I prefer a `getTags()` here, as `count()` is not completely correct: a TagHandler is more than just a collection of tags.

Resources/doc/features/tagging.rst
ddeboer: Todo still open?

Command/InvalidateTagCommand.php
dbu: good idea, adding that in a BC way

EventListener/TagSubscriber.php
ddeboer: Yes, this is unfortunate. It’s due to the `TagHandler` having two (albeit closely related) responsibilities instead of one: both tagging safe responses and invalidating unsafe ones. I don’t see how we can clear this up without splitting the class into two, which may be overkill. 

As for resetting: is our current `TagHandler` suitable for tagging multiple responses? Magically resetting between tag actions may also be confusing if a user wants to tag two responses identically.

Resources/doc/features/tagging.rst
dbu: ups, yes

Handler/TagHandler.php
dbu: or isEmpty? or hasTags?

EventListener/TagSubscriber.php
dbu: i'd say lets wait for a concrete use case and for now keep it with this flaw.

Handler/TagHandler.php
dbu: needs a change on the library though, to add such a method.

dbu:
only open point i can see now is whether we want to have something more elegant than `if ($this->getTagsHeaderValue()) {`

when we clear that, i can squash the commits and we can merge.

Handler/TagHandler.php
ddeboer: `hasTags()` is fine!

dbu: created https://github.com/FriendsOfSymfony/FOSHttpCache/pull/190

we will need to tag 1.3.2 afterwards and this PR depends on 1.3.2 then.

composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "friendsofsymfony/http-cache": "~1.2",
+        "friendsofsymfony/http-cache": "~1.3",
```
dbu: Needs https://github.com/FriendsOfSymfony/FOSHttpCache/pull/190 so tag 1.3.2 and change this to `~1.3,>=1.3.2`

dbu:
depend on fos httpcache 1.3.2 or newer, squashed commits. if travis is happy, this is now ready to merge.

ddeboer:
Small doc issue in the build.

dbu:
now its green.

ddeboer:
Great, thanks!

PR #187, opened 2015-01-15
Fix file path

ericvincenterv:


ddeboer:
That’s right, of course. Thanks!

PR #191, opened 2015-01-18
Fix Travis buid

ddeboer:
I *think* we only needed `minimum-stability: dev` for testing against the dev version of the http-cache library. Unfortunately, all other vendors then got included in their dev versions too. It may be better to work with stable vendor versions (except http-cache) so our tests will suffer less from volatile vendors such as Symfony adding deprecation warnings, for instance. (We catch those warnings for our tests that run against dev version of Symfony thanks to #180.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "symfony/framework-bundle": "~2.3"
     },
     "require-dev": {
+        "friendsofsymfony/http-cache": "~1.2@dev",
         "mockery/mockery": "0.9.*",
         "monolog/monolog": "*",
         "sensio/framework-extra-bundle": "~2.3",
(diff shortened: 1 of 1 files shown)


dbu:
agreed. thanks!

composer.json
```diff
@@ -26,6 +26,7 @@
         "symfony/framework-bundle": "~2.3"
     },
     "require-dev": {
+        "friendsofsymfony/http-cache": "~1.2@dev",
```
stof: don't put the same requirement in both require and require-dev (Composer warns about it). Put the dev flag directly in require (it will only affect the requirement when being in the root package anyway)

ddeboer: Hm yes, but we don't want people to get the unstable version of the library when installing a tagged version of the bundle. Do we then need to keep the minimum-stability?

stof: ``@dev`` in requirements is root-only. It is ignored in dependencies. Only the root package can control stability restrictions

PR #196, opened 2015-02-05
Fixed tagging basic configuration example

lolautruche:
Tags are not updatable under `cache_manager` config, but [directly under root node](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/DependencyInjection/Configuration.php#L65-L75).

[Reference doc](http://foshttpcachebundle.readthedocs.org/en/stable/reference/configuration/tags.html) seems correct though.

Tracked on [eZ community forum](http://share.ez.no/forums/ez-publish-5-platform/fos-http-bundle-and-tagging) :wink:

### DIFF
--- a/Resources/doc/features/tagging.rst
+++ b/Resources/doc/features/tagging.rst
@@ -24,9 +24,8 @@ Then enable tagging in your application configuration:
 .. code-block:: yaml
 
     fos_http_cache:
-      cache_manager:
         tags:
-          enabled: true
+            enabled: true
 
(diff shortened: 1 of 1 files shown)


crevillo:
+1 from me :)

dbu:
ups, indeed. seems we wrote the doc before refactoring this bit - some pre-alpha version had it nested inside i think.

thanks for the fix.

note: the build fail has nothing to do with this change.

lolautruche:
Thanks for merging :smile:

PR #200, opened 2015-02-16
Do not vary response when user identifier and hash are not present in request

joelwurtz:
This fix intend to allow user without session to have cache (but they will all have the same cache as there is no way to distinguish them)

Actually it's sending Cookie and Authorization by default even if we don't have this headers in the request which is not well supported by varnish 4

### DIFF
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -132,7 +132,7 @@ class UserContextSubscriber implements EventSubscriberInterface
             }
         } else {
             foreach ($this->userIdentifierHeaders as $header) {
-                if (!in_array($header, $vary)) {
+                if ($event->getRequest()->headers->has($header) && !in_array($header, $vary)) {
                     $vary[] = $header;
                 }
             }
--- a/Tests/Unit/EventListener/UserContextSubscriberTest.php
+++ b/Tests/Unit/EventListener/UserContextSubscriberTest.php
@@ -156,6 +156,7 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
     {
         $request = new Request();
         $request->setMethod('HEAD');
+        $request->headers->set('X-SessionId', 'session');
 
         $requestMatcher = $this->getRequestMatcher($request, false);
         $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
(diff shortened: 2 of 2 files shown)


dbu:
i fear this is not correct. if i may or may not have a cookie and the response varies on the cookie i have, vary must also be set when there is no cookie header. otherwise varnish will assume that the cached result can be used and ignore that i have a cookie set.

in the varnish configuration, we make varnish cache even when a cookie is set. what exactly is the issue you encounter?

joelwurtz:
My issue is that when my user don't have a session set (Cookie) i don't have cache at all, currently i'm extracting the session id : 

```
if (req.http.cookie) {
    	set req.http.x-userid = ";" + req.http.cookie;
    	set req.http.x-userid = regsuball(req.http.x-userid, "; +", ";");
    	set req.http.x-userid = regsuball(req.http.x-userid, ";(MPSESSID)=", "; \1=");
    	set req.http.x-userid = regsuball(req.http.x-userid, ";[^ ][^;]*", "");
    	set req.http.x-userid = regsuball(req.http.x-userid, "^[; ]+|[; ]+$", "");

	if (req.http.x-userid == "") {
            unset req.http.x-userid;
        }
    }
```

And i'm saying that i want to launch the process of user context hash only if i have this user id

```
if (req.restarts == 0 && req.http.x-userid && (req.method == "GET" || req.method == "HEAD")) {
```

I also have configure fos-http-cache to use this header in the vary header of the response for the user context hash request : 

```
fos_http_cache:
    user_context:
        user_identifier_headers:
            - X-Userid
        role_provider: true
        hash_cache_ttl: 900
```

The problem is when a new user come to this site it does not have any Cookie defining the session and for me all user with no session have the same cache (they are all anonymous and so no specific provider can be set to them)

But the current listener will add `X-UserId` to the response in this case and so on varnish will not cache this page even if there is correct cache headers

joelwurtz:
But i think you are right according to this https://www.varnish-cache.org/trac/wiki/ArchitectureVary, not having a vary will result in a bad cache for people with specific rights.

Another way is maybe to force a value for the user-id by removing the unset in varnish conf ?

dbu:
i am not sure moving the session into a specific header is a good idea. you still should unset the cookies then, to be sure no user specific information is sent to the backend imho...

i guess i would define a hardcoded anonymous user-id. in the symfony HttpCache, there is even a hardcoded user hash for that situation to completely save the backend hash request. we could do that in vcl too, see https://github.com/FriendsOfSymfony/FOSHttpCache/issues/132

ddeboer:
> But the current listener will add X-UserId to the response in this case and so on varnish will not cache this page even if there is correct cache headers

Do you mean your application adds an *empty* X-UserId toe the response? In that case, go with @dbu’s suggestion and just set a hardcoded user id for anonymous users.

joelwurtz:
Like you said @dbu defining a hardcoded value works great : 

I have write the following configuration in order to have cache for anonymous user

```
...

    set req.http.x-userid = "";

    if (req.http.cookie) {
    	set req.http.x-userid = ";" + req.http.cookie;
    	set req.http.x-userid = regsuball(req.http.x-userid, "; +", ";");
    	set req.http.x-userid = regsuball(req.http.x-userid, ";(MPSESSID)=", "; \1=");
    	set req.http.x-userid = regsuball(req.http.x-userid, ";[^ ][^;]*", "");
    	set req.http.x-userid = regsuball(req.http.x-userid, "^[; ]+|[; ]+$", "");
    }

   if (req.restarts == 0 && req.http.x-userid != "" && (req.method == "GET" || req.method == "HEAD")) {
...
``` 

So if user have no session it does not use the user context hash system and instead use the classic scheme of varnish (i have also set below that even if i have cookie i want to cache the response)

And so on it works great with the vary as every anonymous user will have the same value for the x-userid header

PR #201, opened 2015-02-16
adjust example for user context hash url 

### LINKED ISSUE #189: User context documentation tweaks
@dbu (in #188): 
> we should probably add a note in http://foshttpcache.readthedocs.org/en/latest/user-context.html#returning-the-user-context-hash to explain that when using symfony, you should look at the doc for the bundle.

Also, we should note that whereas the lib docs have `user_context_hash.php` as URL, our Symfony docs assume `user-context-hash` (when adding the route). Of course, the Varnish config should then also be changed to `user-context-hash` when using it with Symfony.

dbu:
to be consistent with FOSHttpCache examples and Symfony HttpCache plugins

fix #189

luckily, the URL only occurs in the documentation, not in actual code. the context listener listens only looks for the accept header and matches every url.

### DIFF
--- a/Resources/doc/reference/configuration/user-context.rst
+++ b/Resources/doc/reference/configuration/user-context.rst
@@ -33,7 +33,7 @@ triggered:
 
     # app/config/routing.yml
     user_context_hash:
-        path: /user-context-hash
+        path: /_fos_user_context_hash
 
 .. important::
 


ddeboer:
And again we have Symfony 2.7 tripping over its own deprecation warnings. :cry: This is maybe due to us not ignoring `E_USER_DEPRECATED`, but I’m trying to fix compatibility #202.

ddeboer:
Can you rebase this now #202 has been merged?

dbu:
thanks, rebased.

PR #202, opened 2015-02-16
Better Symfony 2.7/3.0 compatibility

ddeboer:
Adding forwards compatibility for Symfony 2.7/3.0. This makes sense for users already switching over from (deprecated) `TokenStorage` to `SecurityContext` in their applications. This reduces deprecation warnings in our tests. We still need some changes in the bundle config to dynamically inject the token storage instead of the security context service.

### DIFF
--- a/DependencyInjection/Compiler/SecurityContextPass.php
+++ b/DependencyInjection/Compiler/SecurityContextPass.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * In Symfony < 2.6, replace the new security.token_storage service with the  
+ * deprecated security.context service. 
+ */
+class SecurityContextPass implements CompilerPassInterface
+{
+    const ROLE_PROVIDER_SERVICE = 'fos_http_cache.user_context.role_provider';
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has(self::ROLE_PROVIDER_SERVICE)) {
+            return;
+        }
+        
+        if (!$container->has('security.token_storage')) {
+            $definition = $container->getDefinition(self::ROLE_PROVIDER_SERVICE);
+            $definition->replaceArgument(0, new Reference('security.context'));
+        }
+    }
+}
--- a/Tests/Functional/Fixtures/app/config/config.yml
+++ b/Tests/Functional/Fixtures/app/config/config.yml
@@ -5,6 +5,10 @@ framework:
   test: ~
   session:
     storage_id:  session.test_storage
+  # We need to specify templating because of a bug in Symfony: 
+  # see https://github.com/symfony/symfony/issues/13710
+  templating: 
+    engines: ['php']
 
 fos_http_cache:
   cache_control:
--- a/Tests/Unit/UserContext/RoleProviderTest.php
+++ b/Tests/Unit/UserContext/RoleProviderTest.php
@@ -22,8 +22,8 @@ class RoleProviderTest extends \PHPUnit_Framework_TestCase
         $roles = array(new Role('ROLE_USER'));
 
         $token           = \Mockery::mock('Symfony\Component\Security\Core\Authentication\Token\TokenInterface');
-        $securityContext = \Mockery::mock('\Symfony\Component\Security\Core\SecurityContext');
-
+        
+        $securityContext = $this->getTokenStorageMock();
         $securityContext->shouldReceive('getToken')->andReturn($token);
         $token->shouldReceive('getRoles')->andReturn($roles);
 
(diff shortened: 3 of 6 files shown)


UserContext/RoleProvider.php
```diff
@@ -34,10 +35,18 @@ class RoleProvider implements ContextProviderInterface
      * firewall. It is however not valid to call updateUserContext when not in
      * a firewall context.
      *
-     * @param SecurityContextInterface|null $context
+     * @param SecurityContextInterface|TokenStorageInterface $context
      */
-    public function __construct(SecurityContextInterface $context = null)
+    public function __construct($context = null)
     {
+        if ($context && !$context instanceof SecurityContextInterface
```
dbu: lets put the `&& ...` on a separate line to not create the impression those two conditions have a different priority than the last one.

dbu:
:+1: but looks like there is a configuration BC break with 3.0. can we avoid that? otherwise we could have a 3.0 config file and copy that over the default one in the travis setup script.

dbu:
still a problem with some versions...

ddeboer:
Fixed.

DependencyInjection/Compiler/SecurityContextPass.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has(self::ROLE_PROVIDER_SERVICE)) {
+            return;
+        }
+        
+        if (!$container->has('security.token_storage')) {
+            $definition = $container->getDefinition(self::ROLE_PROVIDER_SERVICE);
+            $definition->replaceArgument(0, new Reference('security.context'));
```
ddeboer: We could also alias `security.token_storage` to `security.context` in this case, but it seems to me this messes with Symfony service names too much.

Tests/Functional/Fixtures/app/config/config.yml
```diff
@@ -5,6 +5,8 @@ framework:
   test: ~
   session:
     storage_id:  session.test_storage
+  templating: 
+    engines: ['php']
```
ddeboer: Silly, but unfortunately necessary because of recent changes in Symfony 2.7.x.

DependencyInjection/Compiler/SecurityContextPass.php
dbu: oh no, lets not alias them. that would give people some great WTF...

Tests/Functional/Fixtures/app/config/config.yml
dbu: can you add a comment that this is just to work around a bug in symfony 2.7 (and check if there is a report, or create a report? and then link to that.) so that we remove this again when no longer needed.

dbu:
great! can you add the comment to track the symfony 2.7 bug? otherwise looks fine, lets squash and merge then.

ddeboer:
Added ref to Sf bug.

dbu:
thanks!

DependencyInjection/Compiler/SecurityContextPass.php
stof: if you want an alias, use a private alias in the bundle name space. But don't redefine symfony service ids for that

stof: and shouldn't you set the ``on-invalid="ignore"`` behavior here too, to stay consistent ?

Tests/Functional/Fixtures/app/config/config.yml
```diff
@@ -5,6 +5,10 @@ framework:
   test: ~
   session:
     storage_id:  session.test_storage
+  # We need to specify templating because of a bug in Symfony: 
+  # see https://github.com/symfony/symfony/issues/13710
```
stof: this has been fixed

Tests/Unit/UserContext/RoleProviderTest.php
```diff
@@ -59,4 +58,15 @@ public function testNotUnderFirewall()
...
+    
+    private function getTokenStorageMock()
+    {
+        if (interface_exists('\Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')) {
+            // Symfony >= 2.6
+            return \Mockery::mock('\Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface');
+        }
+        
+        // Symfony < 2.6 compatibility
+        return \Mockery::mock('\Symfony\Component\Security\Core\SecurityContext');
```
stof: should be the interface, not the class

DependencyInjection/Compiler/SecurityContextPass.php
ddeboer: > and shouldn't you set the on-invalid="ignore" behavior here too, to stay consistent ?

@dbu Do you have any idea why `SecurityContext`/`TokenStorage` is an *optional* dependency for the `RoleProvider`? It seems that without context/storage the provider cannot do its work, so why not make it a real dependency?

Tests/Functional/Fixtures/app/config/config.yml
ddeboer: So I noticed. :) I’ll change this back now.

Tests/Unit/UserContext/RoleProviderTest.php
ddeboer: Good point, thanks!

DependencyInjection/Compiler/SecurityContextPass.php
dbu: if i am not mistaken, this needs to be this way because we inject the provider into the hash generator which in turn is injected to an event listener that looks at every request to determine if its a hash lookup request. i would not know of a more elegant solution for this. if you simply make the dependencies required, any request that is not within a firewall will fail.

stof: ``security.context`` is also defined outside a firewall. I guess that making it optional is done for the case where the SecurityBundle is not registered. However, does the listener make sense at all when the security system is not there ? If no, the dependency should be required in the class and the service itself should be removed when the security system is not used

dbu: ah, now that you say this stof: in one of our projects we activate security only in prod environment but not dev. but then we could just not activate the whole hash lookup listener in dev.

so i assume it indeed can be made required if the context is available outside firewalls. but we would need to note that in the changelog as it can have potential to break an update when we hit some edge case like in our project.

PR #207, opened 2015-03-14
add support for a simple etag

### LINKED ISSUE #13: Support for ETag
some sort of mapping (annotations, xml etc) to define how to compute an etag from an object. this might require param converters.

https://github.com/rails/rails/commit/ed5c938fa36995f06d4917d9543ba78ed506bb8d

dbu:
fix #13 
TODO
* [x] adjust full configuration test
* [x] Fix spelling and doc errors


Resources/doc/reference/configuration/headers.rst
```diff
@@ -218,7 +243,12 @@ This value must be a valid input to ``DateTime``.
                     headers:
                         last_modified: "-1 hour"
 
-.. hint::
+.. note::
+
+    Do **not** use last modified and the ETag together, as this only generates
+    confusion. In most situations, you should prefer the ETag method. Last
```
ddeboer: This is very opiniated. Some sources (will try to find them later) consider setting both a best practice.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+1.3.0
+-----
+
+* **2015-03-14** Added support for a simplistic Etag header. This is preferred
```
ddeboer: Let's use ‘simple’ instead of ‘simplistic’ as the later has a rather negative connotation (‘oversimplified’) in English.

ddeboer:
Great to get some ETag scaffolding in!

Resources/doc/reference/configuration/headers.rst
dbu: hm, indeed. i still don't see the use, especially with what we offer (a hardcoded relative last-modified and an md5 hash on the content). maybe i just explain that unless you just want to provide a default last-modified because your logic will only put a last-modified in some situations, you should not activate this.

and on etag tell that if you have a custom strategy for last-modified, activating the md5 etag means that you still need to build the whole page to verify the etag. (well, technically the custom last modified listener could abort the request early with a 304 but it would not know if the etag matches, so there really is no point to it)

shall i do a sidebar that explains what each does and that having both means both must be validated according to the rfc?

dbu:
updated the documentation to make it less opinionated. ok like this?

Resources/doc/reference/configuration/headers.rst
```diff
@@ -218,12 +249,22 @@ This value must be a valid input to ``DateTime``.
...
+.. note::
 
     Setting an arbitrary last modified time allows clients to send
     ``If-Modified-Since`` requests. Varnish can handle these to serve data
     from the cache if it was not invalidated since the client requested it.
 
+    If you only use these simple configurations to add cache validation
+    information, there is no point in adding both a last modified and ETag
+    header. Usually, you should prefer the ETag method as it is not prone to
+    clock skew issues.
```
ddeboer: This is slightly tendentious. There is [RFC 2616 14.25](http://tools.ietf.org/html/rfc2616#section-14.25) which discusses clock skew-related issues with `If-Modified-Since`, but [13.3.4 ‘Rules for When to Use Entity Tags and Last-Modified Dates’](http://tools.ietf.org/html/rfc2616#section-13.3.4) is more nuanced. Perhaps we should at least link to the relevant parts in the RFC so we invite readers to form an opinion of their own.

Resources/doc/features/headers.rst
```diff
@@ -34,7 +34,7 @@ This is an example configuration. For more, see the
                         host: ^login.example.com$
                     headers:
                         cache_control: { public: false, max_age: 0, s_maxage: 0 }
-                        last_modified: "-1 hour"
+                        etag: true
```
ddeboer: Actually, RFC 2616 [recommends](http://tools.ietf.org/html/rfc2616#section-13.3.4) setting *both* ETag and Last-Modified.

Resources/doc/reference/configuration/headers.rst
dbu: indeed. thanks for looking those up! ok, i will relent from giving any opinion and instead just add the links to the rft here. 
i don't understand the why, however. they basically say in 13.3.4 that the client should never change the last modified value in any way, so it becomes semantically the same as an etag to me. maybe twitter will find some answers: https://twitter.com/dbu/status/596594325821198336

Resources/doc/features/headers.rst
dbu: i would prefer to keep the example as is if its ok to you. its just an example anyways.

dbu:
okay, updated the reference. i updated some of our links to RFC 7232 which is more recent, following this [twitter discussion](https://twitter.com/dbu/status/596594325821198336)

andrerom:
should the doc mention how you deal with user right changes? (login/logout) Cause browser will still reuse cache using etag in this case if it is still valid.

dbu:
i think there should be no worry here:

* if you do not use the user context functionality, logged in data will never be cached by the caching proxy, hence etag will never match
* if you use the user context, the content will vary by the credentials. unless the whole page has the exact same content regardless of your rights, that will give a different etag, which in turn means the proxy will not see a cache hit. this actually could be a problem much more likely to happen with last-modified as collisions are more likely.

do i miss something? it would indeed be bad to mix up caches.

andrerom:
ok ,so you generate etag for the response and hence will always need to generate the full response before being able to answer then, missed that part as I saw the discussion on generating etag by object.

dbu:
good point. i added something to the tip to mention this issue. does that make sense to you?

andrerom:
yes, but I'm thinking in direction that permission problem should be handled by FOSHttpCache[Bundle] down the road. It could for instance by convention use `<etag>-<userhash>` as value for "client etag", and validate user hash early as part of proxy (vcl), + strip it before passing to back end.

dbu:
i don't think we need to do this. as the etag is the hash over the full html content, it is either the same content or not. unless you are using a broken cache that mixes variants, but then you have problems even without any etag. if the setup is right, the user hash is used to pick the right variant before comparing the etag, so we are sure the correct tag is checked.

the only situation i could imagine is a client having a cached page and either logged out or changed credentials on the server and sending an if-none-match request. with a custom etag based on just some of the data, there could be the same etag for different variants. then the user would not see that the page changed.

andrerom:
Ok, I was more a fan of generating etag based on object so it can be done early, but it can indeed cause some issues directly because it does not contain user dependant variables. But then I guess when the client checks etag, the proxy will reply not modified based on cache I hope?

I guess this can work, after login client does etag check, varnish varies the cache based on userhash and hence the old etag of anonymous user will not match the variant needed for the user. /cc @lolautruche

dbu:
sure, if you *can* write a listener that decides on the etag early, that is much more efficient.

if you send an etag but keep max-age at 0, the client will revalidate and varnish will look at the correct variant to compare etags.

anyways, if the logged in data is of high privacy concerns, you do not want the client to cache it at all, and send a `no-cache` in the response. otherwise the internet café scenario would be dangerous.

dbu:
@ddeboer do you see any issues with this, or do you agree that its ok like this?

dbu:
ping @ddeboer

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog
 
   **deprecated** `CacheManager::tagResponse` in favor of `TagHandler::addTags`
 * **2015-05-08** Added configuration option for custom proxy client (#208)
+* **2015-05-08** Added support for a simple Etag header in the header configuration (#207)
```
ddeboer: Just wondering: why do you add a date for each entry? Wouldn’t it make more sense to add one date to the release?

dbu: hm, habit i guess. but as the date is when we started working on the feature rather than when it is merged, this is indeed rather pointless. i would not even add a date for the release i think - one can easily check git to know that, or see it on packagist.

i will remove the other dates in master once we merged the PRs, to avoid merge conflicts.

ddeboer: :wink: Let’s do it like that

dbu:
removed the date from the changelog entry.

PR #208, opened 2015-03-14
allow to configure custom proxy client

### LINKED ISSUE #186: Register new proxy clients via service tag
It would be cool to be able to register new proxy client implementations using a compiler pass.

Currently it is only possible to register a new proxy client via. convention, your service must be named as follows:

````
fos_http_cache.test.client.my_custom_client
````

Which breaks namespacing rules.

You could use tags to register new and existing proxy clients:

````
<service id="my_proxy_client" ... />
    <tag name="fos_http_cache.proxy_client"  alias="my_cool_client"/>
</service>
````

### PR DESCRIPTION
**Pull request by dbu**
fix #186

This does not handle the test case section, but then i don't think we need to care about that. The test case would need to know about the caching proxy implementation anyways.

### DIFF
--- a/Resources/doc/reference/configuration/cache-manager.rst
+++ b/Resources/doc/reference/configuration/cache-manager.rst
@@ -2,7 +2,8 @@ cache_manager
 =============
 
 The cache manager is the primary interface to invalidate caches. It is enabled
-by default if a :doc:`Proxy Client <proxy-client>` is configured.
+by default if a :doc:`Proxy Client <proxy-client>` is configured or when you
+specify the ``custom_proxy_client`` field.
 
 .. code-block:: yaml
 
@@ -10,6 +11,7 @@ by default if a :doc:`Proxy Client <proxy-client>` is configured.
     fos_http_cache:
         cache_manager:
             enabled: true
+            custom_proxy_client: ~
             generate_url_type: true
 
 ``enabled``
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: Resources/doc/reference/configuration/proxy-client.rst [hunk 1]
```diff
@@ -1,11 +1,16 @@
...
-service (``fos_http_cache.proxy_client``) that you can use directly.
+The proxy client sends invalidation requests to your caching proxy. The
+:doc:`Cache Manager </reference/cache-manager>` wraps the proxy client and is
+the usual entry point for application interaction with the caching proxy.
+
+You need to configure a client or define your own service for the cache manager
+to work.
+
+The proxy client is also directly available as a service
+(``fos_http_cache.proxy_client.default`` and .varnish or .nginx) that you can use
```
ddeboer: Add backticks around varnish and nginx or perhaps spell them fully for clarity.

dbu:
fixed the doc.

do you agree with naming the option "custom_proxy_client"? i first had proxy_client but while documenting, i realized that this becomes really confusing with a proxy_client section and then a proxy_client field for a service name. and when the proxy_client section is configured, you specify in there which the default client is, it would be wrong to specify the nginx / varnish client service name in the custom_proxy_client field.

dbu:
@ddeboer ping

Resources/doc/reference/configuration/cache-manager.rst
```diff
@@ -10,6 +11,7 @@ by default if a :doc:`Proxy Client <proxy-client>` is configured.
     fos_http_cache:
         cache_manager:
             enabled: true
+            custom_proxy_client: ~
```
ddeboer: What does this mean? Isn’t a service name a requirement for a custom client?

ddeboer:
> do you agree with naming the option "custom_proxy_client"? 

Yes.

Resources/doc/reference/configuration/cache-manager.rst
dbu: the way the configuration works is that this value defaults to null meaning there is no custom client service. if you configure this to a non-null value, that service will be used instead of the default one.

we can't hardcode the client here, because that depends on whether you configure the symfony or varnish or nginx client.

ddeboer: Right!

PR #209, opened 2015-05-08
load purge and refresh handlers by default

### LINKED ISSUE #183: Integrate Symfony HttpCache purge and refresh
Once https://github.com/FriendsOfSymfony/FOSHttpCache/pull/150 is merged, integrate the purge and refresh subscriber and document them.

dbu:
fix #183


dbu:
oh wait, i will adjust the documentation as well.

dbu:
hm, this makes me wonder: we do not have a proxy client for the symfony cache, and no configuration for it. for the user context, we did not need that, but for purge / refresh we need it. i guess we have to add that to the library and add some configuration here?

ddeboer:
Yes, a separate SymfonyHttpCacheClient makes sense. See https://github.com/FriendsOfSymfony/FOSHttpCache/issues/192.

dbu:
this needs https://github.com/FriendsOfSymfony/FOSHttpCache/issues/192 and then configuration for the client, and documentation updates.

dbu:
hm, this reveals an issue with the supported versions of OptionsResolver. the foshttpcache is not properly testing its lowest dependencies.

dbu:
ok so we wait for another version with https://github.com/FriendsOfSymfony/FOSHttpCache/pull/194 and https://github.com/FriendsOfSymfony/FOSHttpCache/pull/193

composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "friendsofsymfony/http-cache": "~1.3,>=1.3.2",
+        "friendsofsymfony/http-cache": "~1.4@dev",
```
dbu: needs to become ~1.4 when 1.4 is released. wait with merging until then.

Resources/config/nginx.xml
```diff
@@ -14,6 +14,7 @@
             <argument>%fos_http_cache.proxy_client.nginx.servers%</argument>
             <argument>%fos_http_cache.proxy_client.nginx.base_url%</argument>
             <argument>%fos_http_cache.proxy_client.nginx.purge_location%</argument>
+            <argument type="service" id="fos_http_cache.proxy_client.nginx.guzzle_client" on-invalid="ignore"/>
```
dbu: i think this is less surprising than what we had before with dynamically adding another argument in the DI extension class.

dbu:
updated. this is ready for review and merge.

ddeboer:
Released lib 1.4.0, so you can adjust dep.

dbu:
and updated to use the stable release of foshttpcache. lets see what travis thinks.

PR #215, opened 2015-06-04
Produce code coverage only once

ddeboer:
This fixes the [variations in code coverage](https://github.com/FriendsOfSymfony/FOSHttpCache/issues/175#issuecomment-81511563) we saw previously, due to hhvm finishing before any of the PHP builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ php:
   - 5.3
   - 5.4
   - 5.5
-  - 5.6
   - hhvm
 
 branches:
@@ -32,6 +31,10 @@ matrix:
       env:
         - SYMFONY_VERSION='2.7.*@dev'
         - FRAMEWORK_EXTRA_VERSION='~3.0'
+    - php: 5.6
+      env:
+        - PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
+        - COVERAGE=true
 
 # only needed for the hackaround for https://github.com/symfony/symfony/issues/12868
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,6 +32,10 @@ matrix:
       env:
         - SYMFONY_VERSION='2.7.*@dev'
         - FRAMEWORK_EXTRA_VERSION='~3.0'
+    - php: 5.6
+      env:
+        - PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
+        - COVERAGE=true
 
 # only needed for the hackaround for https://github.com/symfony/symfony/issues/12868
```
dbu: afaik there is now an official method for this: http://symfony.com/blog/new-in-symfony-2-7-phpunit-bridge 

do you want to try that as well? so we could clean up some of the tricks we did to avoid problems with the deprecation notices.

@WouterJ i thought we are using this in the cmf now, but can't find the right place to point out. can you help us find the example?

dbu:
:+1:

.travis.yml
WouterJ: @dbu just require `symfony/phpunit-bridge` and your ready. Then, fix all deprecations. After this, you're probably left with a few deprecations because:

1. Legacy classes are used to remain BC
2. Things can't be fixed nicely (the usage of `factory-service` in `services.xml` for instance)

For (1), you have to annotate your tests/testcases with `@group legacy` and for (2), you use the env variable `SYMFONY_DEPRECATION_HELPER=weak`.

dbu: thanks a lot wouter! will try this in a separate PR then.

PR #216, opened 2015-06-05
remove hacks to avoid failure due to deprecations

dbu:


ddeboer:
Hm, Scrutinizer now fails because of my #215. The problem is that there’s no way to influence the order of builds on Travis (as far as I’m aware), so when the code coverage build ends up last, we’re way over Scrutinizer’s timeout. I suggest we just raise that timeout to 10 mins. or so.

dbu:
can you change scrutinizer timeout in the configuration panel?

ddeboer:
You can try to set it in this PR:

```yml
external_code_coverage:
        timeout: 600    # Timeout in seconds.
```

ddeboer:
Thanks! Nice to get rid of the hacks. I’ll tweak code coverage in a separate PR.

stof:
Actually, I think jobs are started in the order they are listed in the travis config

PR #217, opened 2015-06-08
Fix typo in symfony service filename.

chtipepere:
Using symfony as proxy_client failed.

### DIFF
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -285,7 +285,7 @@ class FOSHttpCacheExtension extends Extension
 
     private function loadSymfony(ContainerBuilder $container, XmlFileLoader $loader, array $config)
     {
-        $loader->load('symfony-client.xml');
+        $loader->load('symfony.xml');
         foreach ($config['servers'] as $url) {
             $this->validateUrl($url, 'Not a valid web server address: "%s"');
         }
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -427,6 +427,10 @@ class FOSHttpCacheExtension extends Extension
             return 'nginx';
         }
 
+        if (isset($config['symfony'])) {
+            return 'symfony';
+        }
+
         throw new InvalidConfigurationException('No proxy client configured');
     }
 }


chtipepere:
Hmm... Not really sure if this is the issue I'm facing...
I close this PR for the moment.

dbu:
looking at https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/tree/master/Resources/config i think you are correct. and i am quite surprised our tests don't show this problem.

can you please reopen the pull request? i want to merge it.

chtipepere:
I think there are other issues...
I'll re-open the PR and update it.

dbu:
oh, thanks for that one. do you see other issues or is that about it? did you get things to work in your application with this? its all quite new - but obviously we lack some configuration testing here.

chtipepere:
For the moment, I don't know if it's enough.
I'm just trying to use this bundle using symfony reverse proxy on sf2.3
I'll update the PR if I find something else, or add a comment if it's enough.

dbu:
please do a new pull request if you have more things. the fixes you had already are valid and i merged them. please ping when you are good, i will tag a new version then.

chtipepere:
I've finally been able to make the bundle works, but:
- I have to disable tags in config (maybe add this to the documentation)
```yml
    tags:
        enabled:      false
```
Disabling the tags, I've got an other issue:
```php
ServiceNotFoundException: The service "fos_http_cache.command.invalidate_tag" has a dependency on a non-existent service "fos_http_cache.handler.tag_handler".
```
(everything works fine when I comment the service)
I'll dig this issue today

chtipepere:
I'm trying something.
I'll send a new PR in a couple of hours.

dbu:
hm. indeed the symfony cache for now does not support cache tagging. but this is inconvenient for the user.

in symfony cmf, we have some of these options as `true|false|"auto"` and auto means the feature is enabled if it makes sense. that would make sense for the tags. then with symfony or nginx, the whole tag stuff is not enabled, but with varnish it is enabled by default. users can change that to explicit true or false, and in the DI we could catch situations where its configured to `true` but the selected proxy client does not support that feature and give a better error message than a random servicenotfoundexception. wdyt?

chtipepere:
Well.

I've done some stuff to split tag_handler command service from the cache_manager.xml, and I do not have any Exception.
But cache is not invalidate.

Do I miss something, somewhere?

dbu:
you mean you send purge requests to your symfony cache but it does not work? did you set up the HttpCache as explained in the documentation?

chtipepere:
Ok, I've find my problem.
The URL I tried to purge was not well written.
I'll open a new PR with the fix I've made locally to make it works.

dbu:
thanks a lot!

PR #218, opened 2015-06-08
Disable tag invalidation for client with tags disabled.

chtipepere:
I've tried to do the mimimum to make it works with symfony reverse proxy.


Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -102,11 +108,17 @@ symfony
     fos_http_cache:
         proxy_client:
             symfony:
-                servers: 123.123.123.1:6060, 123.123.123.2
-                base_url: yourwebsite.com
+                servers:    123.123.123.1:6060, 123.123.123.2
+                base_url:   yourwebsite.com
```
dbu: lets not change indentation here

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -80,9 +80,15 @@ nginx
                 servers: 123.123.123.1:6060, 123.123.123.2
                 base_url: yourwebsite.com
                 purge_location: /purge
+        tags:
+            enabled:        false
```
dbu: please do not add extra white spaces after `enabled: ` - one is enough.

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -80,9 +80,15 @@ nginx
...
 
 For ``servers`` and ``base_url``, see above.
 
+.. warning::
+
+    By now, tagging only works with Varnish, so ``tags`` has to be explicitly disabled.
```
dbu: do we need to explicitly set it to false? it seems we already have this `auto` option, so if its not specified at all, it should already work, no? or is the `auto` behaving incorrectly?

DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -86,6 +86,9 @@ public function load(array $configs, ContainerBuilder $container)
 
         $container->setParameter($this->getAlias().'.compiler_pass.tag_annotations', $config['tags']['enabled']);
         if ($config['tags']['enabled']) {
+            // only load tag_invalidate if tags are enabled
+            $loader->load('tag_invalidate.xml');
```
dbu: i wonder if we even need to have two separate configuration files. the tag_listener.xml file already contains other services.

@ddeboer what do you think about renaming the file to `cache_tags.xml` or something and have all configuration in there?

Resources/config/cache_manager.xml
```diff
@@ -1,15 +1,14 @@
 <?xml version="1.0" ?>
 
 <container xmlns="http://symfony.com/schema/dic/services"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
```
dbu: not sure about indenting this more. i think 4 spaces is good. not sure what xml conventions in symfony are in general - but the other service files also have just 4 spaces. lets check that separately and if we need to change, we should change consistently.

Resources/doc/reference/configuration/proxy-client.rst
chtipepere: If I do not disable the tags, here's what I faced:
```php
UnsupportedProxyOperationException: HTTP cache does not support BAN requests
```

DependencyInjection/FOSHttpCacheExtension.php
chtipepere: I also think it would be better to put all the "tag related stuff" in one file.
But as it's not my bundle... ;)

Resources/doc/reference/configuration/proxy-client.rst
dbu: ok, this is a bug. by default it should not activate tagging when the proxy client does not support it. i will have a look at that part.

DependencyInjection/FOSHttpCacheExtension.php
ddeboer: I’m in favour of moving all tag-related services into one XML file, as long as we need to either load everything or nothing in that file for the tagging enabled/disabled scenarios, which I think is the case.

DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -88,7 +88,7 @@ public function load(array $configs, ContainerBuilder $container)
         if ($config['tags']['enabled']) {
             // true or auto
             $container->setParameter($this->getAlias().'.tag_handler.header', $config['tags']['header']);
-            $loader->load('tag_listener.xml');
+            $loader->load('tag.xml');
```
dbu: i would opt to call the file cache_tagging.xml to be more explicit - there is several different kind of "tag" in symfony, e.g. service tags.

dbu:
thanks @chtipepere you are quick with updating! one request: can you add some comment when you updated a pull request? i only get notifications from github for comments, not for new commits.

chtipepere:
Pull request has been updated according to your request.

chtipepere:
Checks failed due to a timeout during Scrutinizer job.
The code coverage data was not received within the specified time. Please make sure your third-party service is configured to send code coverage data.

ddeboer:
Fixing that in #221.

Resources/doc/reference/configuration/proxy-client.rst
dbu: #222 will fix that bug. it will however collide with the fixes you are doing in this PR. i propose we check if #222 is green and if so merge that, then i will merge this PR manually and solve the conflict.

chtipepere:
Do I update the PR after #221 will be merged?

Resources/doc/reference/configuration/proxy-client.rst
chtipepere: Alright.

ddeboer:
Both #221 and #222 have been merged, so @dbu please proceed to manually merge this one.

dbu:
see #223

PR #221, opened 2015-06-10
Set Scrutinizer external coverage timeout

ddeboer:
See FriendsOfSymfony/FOSHttpCacheBundle#216.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ branches:
 
 matrix:
   include:
+    - php: 5.6
+      env:
+        - PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
+        - COVERAGE=true
     - php: 5.5
       env: SYMFONY_VERSION='2.3.*'
     - php: 5.5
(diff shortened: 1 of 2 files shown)


stof:
I suggest moving the coverage job higher in the matrix so that it starts earlier, to give it a chance to block scrutinizer for less time

ddeboer:
Will try that, @stof. I think the `matrix:` builds always run later than the `php:` ones, or should I just move the `matrix` block up?

stof:
no, the jobs included manually are indeed added in the matrix after the autogenerated ones. But as more than half our jobs are added in the matrix manually, it will still be better than having it last

.travis.yml
```diff
@@ -32,10 +29,13 @@ matrix:
...
-      env:
-        - PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
-        - COVERAGE=true
+
+php:
+  - 5.3
+  - 5.4
+  - 5.5
+  - 7.0
+  - hhvm
```
stof: changing this is useless (the order of top-level keys is meaningless). and keeping the list of PHP versions on top near the definition of the language makes more sense IMO

ddeboer: That’s what I suspected. Moved it back.

PR #222, opened 2015-06-10
fix DI for tag.enabled=auto and add configuration tests

dbu:
the default `tags.enabled: auto` was ignored and treated as `true` but no validation done whether this will actually work.

this change fixes the problem.

### DIFF
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -84,19 +84,13 @@ class FOSHttpCacheExtension extends Extension
             $loader->load('cache_manager.xml');
         }
 
-        $container->setParameter($this->getAlias().'.compiler_pass.tag_annotations', $config['tags']['enabled']);
         if ($config['tags']['enabled']) {
-            // true or auto
-            $container->setParameter($this->getAlias().'.tag_handler.header', $config['tags']['header']);
-            $loader->load('tag_listener.xml');
-            if (!empty($config['tags']['rules'])) {
-                $this->loadTagRules($container, $config['tags']['rules']);
-            }
-
-            $tagsHeader = $config['tags']['header'];
-            $container->getDefinition($this->getAlias().'.cache_manager')
-                ->addMethodCall('setTagsHeader', array($tagsHeader))
-            ;
+            $this->loadCacheTagging(
+                $container,
+                $loader,
+                $config['tags'],
+                $this->getDefaultProxyClient($config['proxy_client'])
+            );
         }
 
         if ($config['invalidation']['enabled']) {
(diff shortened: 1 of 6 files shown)


DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -84,19 +84,13 @@ public function load(array $configs, ContainerBuilder $container)
...
-            $tagsHeader = $config['tags']['header'];
-            $container->getDefinition($this->getAlias().'.cache_manager')
-                ->addMethodCall('setTagsHeader', array($tagsHeader))
-            ;
+        if (isset($config['proxy_client'], $config['tags'])) {
+            $this->loadCacheTagging(
+                $container,
+                $loader,
+                $config['tags'],
+                $this->getDefaultProxyClient($config['proxy_client'])
```
dbu: tagging would have to work with the default client - tag handler is only wired with the default proxy client (via the cache manager).

DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -84,19 +84,13 @@ public function load(array $configs, ContainerBuilder $container)
...
-            if (!empty($config['tags']['rules'])) {
-                $this->loadTagRules($container, $config['tags']['rules']);
-            }
-
-            $tagsHeader = $config['tags']['header'];
-            $container->getDefinition($this->getAlias().'.cache_manager')
-                ->addMethodCall('setTagsHeader', array($tagsHeader))
-            ;
+        if (isset($config['proxy_client'], $config['tags'])) {
+            $this->loadCacheTagging(
```
ddeboer: Isn’t `$config['tags']` [always set](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/DependencyInjection/Configuration.php#L506)?

DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -84,19 +84,13 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('cache_manager.xml');
         }
 
-        $container->setParameter($this->getAlias().'.compiler_pass.tag_annotations', $config['tags']['enabled']);
         if ($config['tags']['enabled']) {
```
dbu: you are right, this is always defined and the Configuration class pre-validation makes sure that proxy_client is defined when tags are enabled (as its disabled if there is no cache_manager, and there can be no cache_manager if there is no proxy_client)

ddeboer:
Thanks! When we add tagging to the Symfony client, it might make sense to check class interfaces for BanInterface instead if relying on the hard-coded `'varnish'`.

PR #225, opened 2015-06-22
Update match.rst

hbraune:
Underscore was missing in attribute definition.


ddeboer:
Nice catch, thanks! :smile: 

(Build is failing because of docs issue, which I fix in #226.)

Issue #224, opened 2015-06-17
Configuration or annotation invalidation not working in Symfony 2.7.1


benr77:
Hi,

I'm setting up this bundle on Symfony 2.7.1 using the Symfony2 proxy client and of course the Symfony2 HttpCache. If I manually set up invalidation rules in the controller action methods, it works fine and I can see the PURGE requests appearing in the web server logs and the cache is cleared successfully.

However, I cannot get any invalidation to work with the invalidation rules specified either in the YML or as Annotations.

Here is my config

```
fos_http_cache:
  proxy_client:
    symfony:
      servers: 127.0.0.1
      base_url: dev.localhost
  cache_control:
    defaults:
      overwrite: true
    rules:
      -
        match:
          attributes:
            _route: ^booking$
        headers:
          cache_control: { public: true, max_age: 0, s_maxage: 64000 }
          etag: true
          vary: [Accept-Encoding]

      # match everything to set defaults
      -
        match:
          path: ^/
        headers:
          overwrite: false
          cache_control: { public: false, max_age: 0, s_maxage: 0 }
          etag: false
  invalidation:
    rules:
      -
        match:
          attributes: # When hitting these routes
            _route: "booking_update"
        routes:       # Invalidate cache for the following routes
          booking: ~
```

As you can see I'm asking it to invalidate the "booking" route when the "booking_update" route is successfully requested. This appears to do nothing.

I have also tried in the BookingUpdateAction() method the annotation 

`@InvalidateRoute("booking")`

I have the SensioFrameworkExtra bundle installed as per the docs.

I have also modified app/AppCache.php as follows:

```
<?php

require_once __DIR__.'/AppKernel.php';

use Symfony\Component\HttpKernel\HttpKernelInterface;
use FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache;

/**
 * Class AppCache
 */
class AppCache extends EventDispatchingHttpCache
{

}
```

Finally, here is my app_dev.php file

```
<?php

use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\Debug\Debug;

$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
Debug::enable();

//require_once __DIR__.'/../app/AppKernel.php';
require_once __DIR__.'/../app/AppCache.php';

$kernel = new AppKernel('dev', true);
$kernel->loadClassCache();
$kernel = new AppCache($kernel);

[...]

PR #229, opened 2015-07-03
Add a list of default matched status code

tgalopin:
I think it's not very clear for the moment what status code are matched by a rule. I propose to add the list in the documentation, to avoid other problems like https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/224#issuecomment-118378231.

### DIFF
--- a/Resources/doc/reference/configuration/match.rst
+++ b/Resources/doc/reference/configuration/match.rst
@@ -128,8 +128,8 @@ regular expressions.
 
 **type**: ``array``
 
-By defaut, a rule will only match cacheable status codes: 200, 203, 300, 301,
-302, 404 and 410.
+By default, a rule will only match cacheable status codes: 200, 203, 300, 301,
+302, 404 and 410 (as described in the `RFC 7231`_).
 
 `additional_cacheable_status` let you define a list of additional HTTP
 status codes of the response for which to also apply the rule.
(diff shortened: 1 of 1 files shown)


Resources/doc/reference/configuration/match.rst
```diff
@@ -128,8 +128,11 @@ regular expressions.
 
 **type**: ``array``
 
-A list of additional HTTP status codes of the response for which to also apply
-the rule.
+By defaut, a rule will only match cacheable status codes: 200, 203, 300, 301,
```
dbu: can you please link `cacheable status codes` to http://tools.ietf.org/html/rfc7231#page-48 ?

dbu: oh, and [travis noted](https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/69496937) that you have a spelling error: s/defaut/default/

ddeboer: In that RFC, the list is longer: ‘e.g., 200, 203, 204, 206, 300, 301, 404, 405, 410, 414, and 501 in this specification’.

dbu: oh, indeed rfc2616 was more restrictive: http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.4

lets increase the default list but also fix the behaviour that invalidation is possible on any status < 400.

ddeboer:
Merging this, as it correctly reflects the current situation.

Let’s discuss the configurability and up-to-date status codes (for RFC 7231) in #61.

ddeboer:
Thanks @tgalopin!

Issue #69, opened 2014-05-26
RuleMatcher for TagSubscriber, InvalidationSubscriber


dbu:
we should allow to configure a global RuleMatcher in the tag_listener section instead of hardcoding the response codes (see [TagSubscriber $response->isSuccessful](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/TagSubscriber.php#L71)).

i would only allow the match_response and the additional_cacheable_status options i guess, matching the request makes no sense for tags.

ddeboer:
Our scenarios are the following.

1. Cache-Control: should only be set if response is *cacheable* (and request matches0.
2. Invalidation: should only be done if response is *successful*.
3. Tags: should be set if response is *cacheable* and invalidated if response is *successful*.

It makes sense for us to have some sane default for cacheable (RFC 7231) and successful (either 2xx or both 2xx and 3xx). Users should then be able to override these defaults through their configuration. 

Should we have introduce one ResponseMatcher that has `isCacheable()` and `isSuccessful()` or have two separate matchers?

dbu:
lets keep them separate to make them more clear.

wickedOne:
ok, so if i understand correctly you want one or more 'global' match configurations for both request and response which in conjunction define whether or not the response should be cached or be invalidated?

if so, a couple of questions:
- do they overrule the conditions configured in the the ``tags.rules``, do the ``tags.rules`` overrule the 'global' configs, or do they have to be merged upon making that decision (i.e. if you have the ``POST`` method configured in the 'global' rule and the ``PUT`` method in a ``tags.rules.match`` configuration, does the ``request.method`` have to match ``POST``, ``PUT`` or both)?
- do you want a ``ResponseMatcher`` like symfony's ``RequestMatcher`` and if so, what would be the typical properties to match (i assume response code would be one of them)?

dbu:
if i remember correctly, the idea was to use a matcher that would decide whether a response is cacheable or not. if it matches the response, it is cacheable otherwise not. so its not about which tags to apply, but about the status to return.

we brainstormed some additional interfaces in the discussion. we already have a RuleMatcher that adds response handling to the RequestMatcher. but we might want to split that into interfaces that can say whether a response `isCacheable` and whether it `triggersInvalidation`. those could be used in all places we currently use the request::isSuccessful and response status code checks.

wickedOne:
[...]

PR #230, opened 2015-07-10
Fix invalidation not being triggered for some status codes

### LINKED ISSUE #224: Configuration or annotation invalidation not working in Symfony 2.7.1

ddeboer:
Fix #224. Invalidation was not triggered when the controller returned a response with status codes such as 204 or 302. A more elaborate solution, which involves creating separate, configurable response matchers (as well as request matchers), as described in #69, should be done in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.3.2
+-----
+
+* Fixed some status codes (such as 204 and 302) not triggering invalidation.
+
 1.3.1
 -----
 
--- a/EventListener/InvalidationSubscriber.php
+++ b/EventListener/InvalidationSubscriber.php
@@ -89,7 +89,9 @@ class InvalidationSubscriber extends AbstractRuleSubscriber implements EventSubs
         $response = $event->getResponse();
 
         // Don't invalidate any caches if the request was unsuccessful
-        if ($response->isSuccessful()) {
+        if ($response->getStatusCode() >= 200
+            && $response->getStatusCode() < 400
+        ) {
             $this->handleInvalidation($request, $response);
         }
 
@@ -144,7 +146,8 @@ class InvalidationSubscriber extends AbstractRuleSubscriber implements EventSubs
     /**
      * Handle the invalidation annotations and configured invalidators.
      *
-     * @param Request $request
+     * @param Request  $request
+     * @param Response $response
      */
     private function handleInvalidation(Request $request, Response $response)
     {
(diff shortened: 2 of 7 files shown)


dbu:
thanks a lot!

Issue #176, opened 2014-11-30
Fix Symfony deprecation warnings


ddeboer:
Tests are failing because deprecation warnings were added – probably since the SymfonyCon hackday. :)

How are we going to cope with these deprecation warnings? They can only be fixed by relying on classes that were added in 2.5. Should we:
- maintain two branches (rather not)
- or add checks in our code to call the right classes depending on the Symfony version?

I guess when next LTS is out (May 2015) we can require users to upgrade from 2.3 to that.

PR #239, opened 2015-09-29
Test against Symfony 2.8

### LINKED ISSUE #238: The parent definition "fos_http_cache.test.client.abstract" ... does not exist.
````
The parent definition "fos_http_cache.test.client.abstract" defined for definition "fos_http_cache.test.client.varnish" does not exist.
````

The `fos_http_cache.test.client.varnish` service extends this service, but that service is only loaded when the `test` definitions are loaded (i.e. when the `test` key is included in the config).

### PR DESCRIPTION
**Pull request by ddeboer**
Test against Symfony 2.8. Fix #238.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ php:
   - 5.3
   - 5.4
   - 5.5
+  - 5.6
   - 7.0
   - hhvm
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by ddeboer** [comment]
Okay, #238 is resolved but we’re getting another error now.

File: .travis.yml [hunk 1]
```diff
@@ -45,7 +50,7 @@ before_script:
   - sudo pip install -r Resources/doc/requirements.txt
 
 script:
-  - phpunit ${PHPUNIT_FLAGS}
+  - SYMFONY_DEPRECATIONS_HELPER=weak phpunit ${PHPUNIT_FLAGS}
```
ddeboer: And [again](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/176#issuecomment-65728870) we were getting failed builds because of deprecation warnings caused by internal Symfony code. @stof Is this the recommended way to solve this problem?

ddeboer:
Ready for review.

dbu:
thanks! and yes this looks correct to me. we want to get rid of the warnings at some point to get ready for symfony 3.

to make our lives not too hard i think we should just bump the minimum version to 2.7 to not bother too much with code to work with symfony 2.3 not having all of the things that replace the things that are deprecated in 2.7/2.8.

PR #243, opened 2015-10-15
[SymfonyCache] Remove event classes from Symfony's class cache to fix #242

### LINKED ISSUE #242: Fatal error: Cannot redeclare class Symfony\Component\EventDispatcher\Event in POST requests
Since I have installed FOSHttpCacheBundle I get a strange error on POST requests (when submitting any form): 
```
Fatal error: Cannot redeclare class Symfony\Component\EventDispatcher\Event in C:\projects\acme\app\cache\dev\classes.php on line 2136

Call Stack:
    0.0020     136568   1. {main}() C:\projects\acme\web\app_dev.php:0
    0.0380    1908496   2. FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache->handle() C:\projects\acme\web\app_dev.php:34
    0.0380    1908952   3. Symfony\Component\HttpKernel\HttpCache\HttpCache->handle() C:\projects\acme\vendor\friendsofsymfony\http-cache-bundle\SymfonyCache\EventDispatchingHttpCache.php:122
    0.0390    1928528   4. FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache->invalidate() C:\projects\acme\vendor\symfony\symfony\src\Symfony\Component\HttpKernel\HttpCache\HttpCache.php:206
    0.0410    1948712   5. Symfony\Component\HttpKernel\HttpCache\HttpCache->invalidate() C:\projects\acme\vendor\friendsofsymfony\http-cache-bundle\SymfonyCache\EventDispatchingHttpCache.php:161
    0.0410    1948736   6. Symfony\Component\HttpKernel\HttpCache\HttpCache->pass() C:\projects\acme\vendor\symfony\symfony\src\Symfony\Component\HttpKernel\HttpCache\HttpCache.php:275
    0.0410    1949016   7. Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache->forward() C:\projects\acme\vendor\symfony\symfony\src\Symfony\Component\HttpKernel\HttpCache\HttpCache.php:258
    0.0410    1949176   8. Symfony\Component\HttpKernel\Kernel->boot() C:\projects\acme\vendor\symfony\symfony\src\Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache.php:56
    0.0410    1949240   9. Symfony\Component\HttpKernel\Kernel->doLoadClassCache() C:\projects\acme\app\bootstrap.php.cache:2408
    0.0430    2034240  10. Symfony\Component\ClassLoader\ClassCollectionLoader::load() C:\projects\acme\app\bootstrap.php.cache:2573
    0.1000    4173632  11. require_once('C:\projects\acme\app\cache\dev\classes.php') C:\projects\acme\vendor\symfony\symfony\src\Symfony\Component\ClassLoader\ClassCollectionLoader.php:91
```
Now, when I delete the cache folder via filesystem (**not** with app/console cache:clear) the next POST request works, but all next POST requests are failing again with the same error message.
But when I execute `app/console cache:clear` after removing the cache folder it seem to work on all requests.

I have already figured out that on POST requests the `Symfony\Component\EventDispatcher\Event` class is loaded via autoloading when an instance of `CacheEvent` is created in `FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache::invalidate`. Later the Symfony HttpKernel boots and loads the class cache (the *classes.php* in the cache folder) which contains the Event class, too. This results in the mentioned redeclaration error.

What confuses me, is that the Event class is not compiled into classes.php when `app/console cache:clear` is executed after removing the cache folder.

Does anyone has a clue about this issue?

naitsirch:
Symfony's class cache (*cache/<env>/classes.php*) compiles classes that are loaded on each request.

But when using `EventDispatchingHttpCache` there may occure errors, when using Symfony's HttpCache, because EventDispatchingHttpCache loads some classes (`Symfony\Component\EventDispatcher\Event`) individually via autoload and later when the HttpKernel is bootet in a subrequest and the class cache is loaded, PHP runs into a fatal error because of redeclaration of the class `Symfony\Component\EventDispatcher\Event` (as you can see in #242).

To fix this issue I have added an compiler pass to remove the mentioned classes from the cache class map before the cache is built. This only happens if cache invalidation is used (as it only happens in that case, afaik).

If there are any objections/annotations to this PR, please let me know.

### DIFF
--- a/DependencyInjection/Compiler/RemoveClassesFromCachePass.php
+++ b/DependencyInjection/Compiler/RemoveClassesFromCachePass.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Compiler pass to remove some classes from Symfony's class cache to prevent
+ * from redeclaration in early cache lookup phase
+ * (see https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/242).
+ */
+class RemoveClassesFromCachePass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasParameter('fos_http_cache.invalidation.enabled')) {
+            return;
+        }
+
+        $frameworkExt = $container->getExtension('framework');
+        $classes = $frameworkExt->getClassesToCompile();
+
+        // remove classes
+        $classesToRemove = array(
+            'Symfony\\Component\\EventDispatcher\\Event',
+            'Symfony\\Component\\HttpKernel\\Event\\KernelEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\FilterControllerEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\FilterResponseEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseForControllerResultEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent',
+        );
+        foreach ($classesToRemove as $className) {
+            $index = array_search($className, $classes);
+            array_splice($classes, $index, 1);
+        }
+
+        // overwrite array for class cache via reflection (no other way)
+        $reflClass = new \ReflectionClass('Symfony\\Component\\HttpKernel\\DependencyInjection\\Extension');
+        $reflProp = $reflClass->getProperty('classes');
+        $reflProp->setAccessible(true);
+        $reflProp->setValue($frameworkExt, $classes);
--- a/DependencyInjection/Compiler/RemoveClassesFromCachePass.php
+++ b/DependencyInjection/Compiler/RemoveClassesFromCachePass.php
@@ -53,3 +53,5 @@ class RemoveClassesFromCachePass implements CompilerPassInterface
         $reflProp = $reflClass->getProperty('classes');
         $reflProp->setAccessible(true);
         $reflProp->setValue($frameworkExt, $classes);
+    }
+}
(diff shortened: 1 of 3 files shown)


stof:
There could be a simpler fix for that: force loading the necessary event classes in the build method of the bundle (the classes being used by the kernel wrapper), to be sure that these classes do not get included in the compiled file, as they will already have been loaded.
This is necessary because the list of classes may be dumped in cases where the event was not triggered first.

stof:
your own fix is very fragile:

- any other bundle adding an event class would trigger the same issue again
- any new event class added in FrameworkBundle would trigger it too
- it relies on Reflection to access private state in Symfony, and there is strictly no BC guarantee on that (even in patch releases)

naitsirch:
Thanks for your hint @stof. Can you give me a small example how to *force loading the necessary event classes in the build method of the bundle*?

naitsirch:
Do you mean to simply add a 'new \FOS\HttpCache\SymfonyCache\CacheEvent()' in the bundle's build method?

DependencyInjection/Compiler/RemoveClassesFromCachePass.php
```diff
@@ -0,0 +1,57 @@
...
+
+        // remove classes
+        $classesToRemove = array(
+            'Symfony\\Component\\EventDispatcher\\Event',
+            'Symfony\\Component\\HttpKernel\\Event\\KernelEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\FilterControllerEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\FilterResponseEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseForControllerResultEvent',
+            'Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent',
```
dbu: if the class cache is relevant, this might be an expensive thing. how do other bundles that have events solve this issue?

stof:
@naitsirch just a ``class_exists`` call to trigger the autoloading is enough. No need to instantiate it

DependencyInjection/Compiler/RemoveClassesFromCachePass.php
stof: @dbu the issue of this bundle is that it has an event before reaching the AppKernel (in its AppCache). Other bundles don't

dbu: ah, thanks. yeah now i get it.

dbu:
so we need a `class_exists` on each event class our AppCache uses in our extension class?

naitsirch:
@dbu no, it's enough to call `class_exists('FOS\\HttpCache\\SymfonyCache\\CacheEvent');` I'll create a second PR, soon ;)

naitsirch:
Closed in favor of #244

PR #245, opened 2015-11-07
Use decodeURIComponent to decode cookie before parsing it

skonsoft:
When we use this function as it is, we get a javascript error saying:

```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
	var a=JSON.parse(b);console.log(a);...
```
In the server side, json_encode encrypt the cookie in HTTP, so to parse the cookie we should decrypt that can by using decodeURIComponent javascript function.

### DIFF
--- a/Resources/doc/features/helpers/flash-message.rst
+++ b/Resources/doc/features/helpers/flash-message.rst
@@ -49,7 +49,7 @@ to only show the flash message once. Something along these lines:
             return;
         }
 
-        var flashes = JSON.parse(cookie);
+        var flashes = JSON.parse(decodeURIComponent(cookie));
 
         // show flashes in your DOM...
 


dbu:
oh, did you get the flash message handler to work? that sounds great!

if you could read over the documentation section and clarify it, that would be great. we ported it from the old LiipCacheControlBundle but found nobody actually using it... i am not even 100% sure i understand the idea: we respond with a cookie during the redirect message after a POST that can't be cached anyways? so that the GET request after the redirection can be cached, and a flash message displayed through javascript? if i have that correctly, we should update the documentation to explain this process.

skonsoft:
I don't know what the problem exactly, but here the code that i used to show flashes using the javascript and cookies:

```javascript
var flashDivId = '#flashes',
        flashCookieName = 'flashes'; // fos_http_cache.flash_message.name

function getCookie(cname)
{
    var name = cname + "=";
    var ca = document.cookie.split(';');
    for (var i = 0; i < ca.length; i++) {
        var c = ca[i].trim();
        if (c.indexOf(name) == 0) {
            return c.substring(name.length, c.length);
        }
    }

    return false;
}

function showFlash()
{
    var cookie = getCookie(flashCookieName);

    if (!cookie) {
        return;
    }

    cookie = cookie.replace(/\+/g, '%20'); //replace the + by an http encoded element
    var flashes = jQuery.parseJSON(decodeURIComponent(cookie));

    var html = '';

    for (var key in flashes) {
        if (key === 'length' || !flashes.hasOwnProperty(key)) {
            continue;
        }
        var msg = flashes[key];
        html = '<div class="alert alert-' + key + '">';
        html += '<i class="fa fa-info-circle"></i>';
        html += msg;
        html += '</div>';
    }

// show flashes in your DOM...
    $(flashDivId).html(html);

    //Remove the cookie
    $.removeCookie(flashCookieName, {path: '/'});
}

// register showFlash on the page ready event.

$(function () {
    showFlash();
});

```

I hope this help you to fix this issue.

dbu:
thanks. added some of this in #248

PR #247, opened 2015-11-10
Enable prioritising tagged definitions.

### LINKED ISSUE #246: Priority for ContextProviders
It is not possible to replace or prioritise the order in which context providers are supplied, meaning that it is not currently possible to consistently remove or add headers to the existing context hash.

antoligy:
Closes #246.

### DIFF
--- a/DependencyInjection/Compiler/HashGeneratorPass.php
+++ b/DependencyInjection/Compiler/HashGeneratorPass.php
@@ -56,8 +56,6 @@ class HashGeneratorPass implements CompilerPassInterface
             $providers[] = new Reference($id);
         }
 
-        if (!empty($providers)) {
-            $definition->addArgument($providers);
-        }
+        $definition->addArgument($providers);
     }
 }
--- a/DependencyInjection/Compiler/HashGeneratorPass.php
+++ b/DependencyInjection/Compiler/HashGeneratorPass.php
@@ -34,13 +34,25 @@ class HashGeneratorPass implements CompilerPassInterface
 
         $definition = $container->getDefinition('fos_http_cache.user_context.hash_generator');
 
-        $providers = array();
-        foreach ($container->findTaggedServiceIds(self::TAG_NAME) as $id => $parameters) {
-            $providers[] = new Reference($id);
+        $prioritisedTags = array();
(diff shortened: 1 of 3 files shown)


DependencyInjection/Compiler/HashGeneratorPass.php
```diff
@@ -34,15 +34,30 @@ public function process(ContainerBuilder $container)
...
+
         $providers = array();
-        foreach ($container->findTaggedServiceIds(self::TAG_NAME) as $id => $parameters) {
+        foreach ($prioritisedProviders as $id) {
             $providers[] = new Reference($id);
         }
 
-        if (!count($providers)) {
-            throw new InvalidConfigurationException('No user context providers found. Either tag providers or disable fos_http_cache.user_context');
+        if (!empty($providers)) {
```
dbu: how could providers be empty now? you kept the sanity check with `count` above, so i don't think we need this check.

dbu:
looks good, and i see that this makes sense. i have one comment on the code.

can you please also add a short section to the documentation (in Resources/doc) that explains this feature?

DependencyInjection/Compiler/HashGeneratorPass.php
antoligy: This was in response to Scrutinizer's "high" rated issue wherein apparently (???) it was not always possible to fulfil $providers.

antoligy:
Certainly. :smile:

DependencyInjection/Compiler/HashGeneratorPass.php
dbu: then that looks like a scrutinizer bug, right?

antoligy: I think so, should I remove it?

dbu: yes, lets remove it.

antoligy:
Just added some docs, and have responded to feedback.

antoligy:
Any further feedback?

dbu:
thanks a lot! i squashed the commits and merged in b24ab0cfdb117d88ff83660b70b03329c2c1b813

PR #248, opened 2015-11-10
clarify the javascript side for the flash listener

dbu:
see #245

@skonsoft does this make sense?

### DIFF
--- a/Resources/doc/spelling_word_list.txt
+++ b/Resources/doc/spelling_word_list.txt
@@ -10,3 +10,4 @@ yaml
 invalidator
 ETag
 templating
+Javascript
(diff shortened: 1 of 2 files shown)


Resources/doc/spelling_word_list.txt
```diff
@@ -10,3 +10,4 @@ yaml
 invalidator
 ETag
 templating
+Javascript
```
ddeboer: Should be `JavaScript`.

ddeboer:
:+1:

Resources/doc/spelling_word_list.txt
dbu: sorry, missed that. fixed in f2623ce

PR #250, opened 2015-11-16
drop php 5.3 support

dbu:
PHP 5.3 is past its end of live date. 

And builds fail with out of memory problems...
https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/91344571

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
+        "php": ">=5.4.0",
         "friendsofsymfony/http-cache": "~1.4",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -48,7 +48,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.3.x-dev"
+            "dev-master": "2.0.x-dev"
```
dbu: the release 1.3.4 should be the last in the 1.x line. master already has a new feature with the priorities, but i missed to update the branch alias.

ddeboer:
Goodbye 5.3! :wave:

PR #235, opened 2015-07-31
update to use 2.0 of FOSHttpCache

dbu:
- Updating to FOSHttpCache 2.0
- Starting 2.0 development
- Adjust documentation for support of custom ttl in the library.

TODO
- [x] We need something to configure the http client: https://github.com/php-http/HttplugBundle/
- [x] figure out configuration for http clients in the proxy clients

fix #219

PR #255, opened 2015-12-04
Test against Symfony 3.0

ddeboer:


dbu:
should we do this against the 1.3 branch? it seems master atm does not install anymore due to changes within php-http. i propose we don't try to solve that in here, as we first need to wrap up https://github.com/FriendsOfSymfony/FOSHttpCache/pull/257 and #235

dbu:
why does this fail? https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/96291586

`Twig_Error_Syntax: Unknown "asset" function in "@Twig/Exception/exception_full.html.twig" at line 4.`

we do have the TwigBundle as we depend on symfony/symfony, and we do instantiate the bundle it in the kernel that is used by the functional test. is TwigBundle in 3.0.0 broken?

dbu:
ups. wrong PR. was talking about #268. closing this one in favor of #268

PR #256, opened 2015-12-04
Configurable ExpressionLanguage

### LINKED ISSUE #252: Configurable ExpressionLanguage class
It would be nice to be able to define your own ``ExpressionLanguage`` class.
This would enable additional / custom functionality for tagging with configuration rules.

wickedOne:
Configuration option for a custom expression language service.
Fixes #252


Resources/doc/reference/configuration/invalidation.rst
```diff
@@ -20,7 +19,25 @@ routes when some other routes are requested.
...
+.. include:: ../../includes/expression-language.rst
+
+Your custom expression functions can then be used in the ``@InvalidateRoute`` :ref:`annotations<invalidateroute>`.
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    services:
+        acme.expression_language:
+            class: Acme\TestBundle\Component\ExpressionLanguage
```
dbu: lets use the new best practices for examples and put this in AppBundle. maybe `AppBundle\ExpressionLanguage\MyExpressionLanguage` ?

Resources/doc/reference/configuration/tags.rst
```diff
@@ -29,6 +29,25 @@ incompatible proxies:
...
+
+Your custom expression functions can then be used in both the ``tag_expressions``
+section of the tag configuration and ``@tag`` :ref:`annotations<tag>`.
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    services:
+        acme.expression_language:
+            class: Acme\TestBundle\Component\ExpressionLanguage
```
dbu: should we make the code example an include as well? or actually, i would just say that you define a custom service without actually showing how to define the service, its a bit much details

Resources/doc/reference/configuration/invalidation.rst
```diff
@@ -20,7 +19,25 @@ routes when some other routes are requested.
                         villain_details:     # e.g., /villain/{id}
                             ignore_extra_params: false    # Defaults to true
 
-.. include:: /includes/enabled.rst
+.. include:: ../../includes/enabled.rst
```
dbu: i think we had absolute paths on purpose. they are absolute in regard to the documentation root, so sphinx should be able to work with them. did you run into an issue?

dbu:
great, thanks a lot!
my only change requests are about the documentation, code wise this looks good to me. @ddeboer do you agree?

Resources/doc/reference/configuration/invalidation.rst
wickedOne:
```bash
reST markup error:
/vagrant/vendor/friendsofsymfony/http-cache-bundle/Resources/doc/reference/configuration/invalidation.rst:43: (SEVERE/4) Problems with "include" directive path:
IOError: [Errno 2] No such file or directory: 'reference/configuration/includes/enabled.rst'.
```
might be my sphinx version though ``Running Sphinx v0.6.6``

wickedOne:
just let me know whether i should revert back to absolute paths and i'll amend the suggested changes

ddeboer:
Yeah, please keep the paths absolute. :wink: If we do need to change them, I prefer to do that in a separate PR.

Resources/doc/reference/configuration/invalidation.rst
dbu: i am running sphinx-build -v 1.2.2 so maybe you would need to upgrade.
did you run from inside the doc/ directory and type `make html`?
otherwise sphinx will not be configured correctly

wickedOne:
ok here you go, suggested changes amended and pushed…

dbu:
something is wrong with php 7 and the phpunit version. but that has nothing to do with the PR, i will merge this. thanks!

PR #258, opened 2015-12-04
[proxy client] guzzle client config fix

### LINKED ISSUE #257: [proxy client] guzzle client configuration
while working on https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/256 today i noticed that the configuration of the custom ``guzzle_client`` for the various proxy clients is done by 
```php
if (!empty($config['guzzle_client'])) {
    $container->setParameter($this->getAlias().'.proxy_client.nginx.guzzle_client', $config['guzzle_client']);
}
```
which, as far as i know, will raise a
```bash
The service "fos_http_cache.proxy_client.nginx" has a dependency on a non-existent service "fos_http_cache.proxy_client.nginx.guzzle_client"
```

shouldn't that be as following?
```php
if (!empty($config['guzzle_client'])) {
    $container->setAlias(
        $this->getAlias().'.proxy_client.nginx.guzzle_client',
        $config['guzzle_client']
    );
}
```

wickedOne:
fixes #257 
that is, if i'm right about that one…

### DIFF
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -263,8 +263,12 @@ class FOSHttpCacheExtension extends Extension
         }
         $container->setParameter($this->getAlias().'.proxy_client.varnish.servers', $config['servers']);
         $container->setParameter($this->getAlias().'.proxy_client.varnish.base_url', $baseUrl);
+
         if (!empty($config['guzzle_client'])) {
-            $container->setParameter($this->getAlias().'.proxy_client.varnish.guzzle_client', $config['guzzle_client']);
+            $container->setAlias(
+                $this->getAlias().'.proxy_client.varnish.guzzle_client',
+                $config['guzzle_client']
+            );
         }
     }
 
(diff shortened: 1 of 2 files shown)


dbu:
hm, i think you are right. surprising that this never has been a problem - maybe people where just always using their custom guzzle clients? @ddeboer or am i missing something?

there is one unit test that needs to be adjusted: https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/94924465

note that for 2.0, we will refactor the library and bundle to use the Httplug abstraction, so this code will go away. but its already the same in the 1.3 branch. can you maybe create a new pull request against 1.3? (create a branch out from 1.3, then cherry-pick this commit into it, fix the test and do a new PR selecting 1.3 as the target branch)

wickedOne:
i noticed my pr had failed, but decided to wait with fixing it until it was confirmed this was actually an issue… will try and get to that later this week

wickedOne:
@dbu would it be ok to re-open this pr and have it merged into the master branch as well?

even though the code will be removed for version 2.0, it currently is a broken feature (and i kind of need both ``expression_language`` and ``guzzle_client`` configuration options :sunglasses: )

dbu:
i want to merge 1.3 into master once we have the fix there. otherwise we
get a confusing git history.

PR #261, opened 2015-12-09
Fix dependency on a non-existent service "security.context"

Tobion:
The security part is optional in the role provider class, which is why `security.token_storage` is also defined with `on-invalid=ignore`.
As we have security disabled in dev env, neither security.token_storage nor security.context exists, but it still tries to replace the dependency and loses on-invalid=ignore. Then it was complaining, that the service is missing.

### DIFF
--- a/DependencyInjection/Compiler/SecurityContextPass.php
+++ b/DependencyInjection/Compiler/SecurityContextPass.php
@@ -22,7 +22,7 @@ use Symfony\Component\DependencyInjection\Reference;
 class SecurityContextPass implements CompilerPassInterface
 {
     const ROLE_PROVIDER_SERVICE = 'fos_http_cache.user_context.role_provider';
-    
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


Tobion:
@ddeboer @dbu would be nice to merge this and it prevents us to upgrade to symfony 2.8

dbu:
thanks. the build fails for some versions, but this also happens with the main 1.3 branch and seems not related. i think we fixed that one on master but not in 1.3

i will have a look and try to fix that before tagging a patch release. for now you should be able to use `1.3.*@dev`

Tobion:
The versioning for this bundle is messed up. 
- Master branch is allegedly version 2.x: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/composer.json#L51
- But some things have only been merged into master ONLY and tagged as 1.3.4:  #239

Tobion:
So please backport #239 to branch 1.3 (and/or change the branch alias)

dbu:
argh. master is indeed about refactorings for 2.0.

i cherry-picked the few commits (all from #239) to the 1.3 branch and if
travis is green will tag 1.3.5

dbu:
thanks for realizing that. seems we missed the split with the 1.3 branch
in that cleanup.

dbu:
also merged 1.3 into master to get your fix there.

Tobion:
Looks good, only a memory problem on PHP 5.3

ddeboer:
Thanks @Tobion for pointing out our mistake. Just tagged a fixed 1.3.5 version. I considered it more important to get the update out as soon as possible than to first fix the 5.3 memory problem on Travis.

Tobion:
Thanks

PR #262, opened 2015-12-09
Removed over-zealous configuration checks

benr77:
With these checks removed you can now specify `enabled: false` on invalidation or tagging configuration sections even if you have previously defined some rules for that section.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -79,10 +79,6 @@ class Configuration implements ConfigurationInterface
                 })
             ->end()
             ->validate()
-                ->ifTrue(function ($v) {return $v['tags']['rules'] && !$v['tags']['enabled'];})
-                ->thenInvalid('You need to enable the cache_manager and tags to use rules.')
-            ->end()
-            ->validate()
                 ->ifTrue(function ($v) {return $v['invalidation']['enabled'] && !$v['cache_manager']['enabled'];})
                 ->then(function ($v) {
                     if ('auto' === $v['invalidation']['enabled']) {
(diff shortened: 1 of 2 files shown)


dbu:
:+1:

can you please check the ConfigurationTest::testTagRulesNotEnabled and testInvalidationRulesNotEnabled? those now fail because of wrong assumptions. we should make them assume the new behaviour: https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/95925849

are things correctly ignored later on, in the extension class?

benr77:
The extension class only loads the rules if the invalidation or tagging is enabled, so yes I think things are handled correctly. Should the two tests be removed entirely? To correspond with the removal of the actual validation rules?

dbu:
i think we should keep them, but verify that no rules are stuffed into the container if the system is not activated

benr77:
But that is not the behaviour - from what I understand the rules currently cannot be overridden or removed once they are defined, so if you override the enabled setting the rules will already be in the container.

So it would first need to be possible to overwrite any existing rules if enabled is defined as false.

I might be getting mixed up with this - I have very little experience with bundle configuration...

dbu:
ah, actually this is the ConfigurationTest. so yes, you can delete them as they test a case that indeed no longer exists. the configuration will return information, but then the extension class will ignore it. (i was thinking we test what the extension puts into the container definition, but i was wrong...)

dbu:
thanks! be aware that the master branch of this bundle (2.0) is quite a bit a moving target atm. we will switch to Httplug at some point. if you need to use it in your project, i recommend requiring a specific commit id of this and the foshttpcache library to not have unexpected errors when updating dependencies, until we get to the point of RCs

PR #263, opened 2015-12-10
build 5.3 in prefer-lowest

dbu:
lets hope the out of memory error on 5.3 goes away with this. and we where missing the --prefer-lowest build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,12 +34,15 @@
         "symfony/expression-language": "~2.4",
         "symfony/monolog-bundle": "~2.3",
         "polishsymfonycommunity/symfony-mocker-container": "~1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "0.*"
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7.4"
     },
     "suggest": {
         "sensio/framework-extra-bundle": "For Tagged Cache Invalidation",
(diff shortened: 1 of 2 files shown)


dbu:
still results in out of memory. should we simply drop php 5.3 from the matrix? its dead anyways.

ddeboer:
Agreed. Let’s keep 5.3 in `composer.json` for now though, so we don’t introduce a BC break for users that are still on 5.3 (if there are any).

dbu:
updated.

technically upping the requirement is not a BC break, it just means
people on 5.3 can't update any longer. but i agree, we drop it in 2.0
which is already in progress.

composer.json
```diff
@@ -34,13 +34,16 @@
...
-        "matthiasnoback/symfony-dependency-injection-test": "0.*"
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7.4"
     },
     "suggest": {
         "sensio/framework-extra-bundle": "For Tagged Cache Invalidation",
         "symfony/expression-language": "For Tagged Cache Invalidation",
         "symfony/console": "To send invalidation requests from the command line"
     },
+    "conflict": {
+        "twig/twig": "<2.0.0"
```
dbu: we use Twig_SimpleFunction

PR #268, opened 2015-12-11
Symfony 3

dbu:
replaces #255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,15 +53,21 @@ matrix:
         - FRAMEWORK_EXTRA_VERSION='~3.0'
     - php: 5.6
       env:
-        - SYMFONY_VERSION='2.8.*@dev'
+        - SYMFONY_VERSION='2.8.*'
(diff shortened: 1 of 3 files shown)


dbu:
why does this fail? https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/96291586

`Twig_Error_Syntax: Unknown "asset" function in "@Twig/Exception/exception_full.html.twig" at line 4.`

we do have the TwigBundle as we depend on symfony/symfony, and we do instantiate the bundle it in the kernel that is used by the functional test. is TwigBundle in 3.0.0 broken?

stof:
@dbu in 3.0, the assets component is not enabled by default anymore in FrameworkBundle. Add ``assets: ~`` in the config file (see https://github.com/symfony/assetic-bundle/blob/master/Tests/TestKernel.php#L39 for a way to do it if you need to support older versions of Symfony too)

dbu:
thanks @stof , exactly what I needed. you are simply amazing!

PR #271, opened 2016-01-07
include request object in match_response

### LINKED ISSUE #270: include request object in match_response
would it be ok for me to add the request object in the ``match_response`` evaluation in the [RuleMatcher](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Http/RuleMatcher.php#L62-L64)?

i'm currently working on an api which returns a ``201`` response code for successful ``PUT`` requests.
cleary i don't want to add the ``201`` to the ``additional_cacheable_status``, so when adding the request to the evaluation i could do something like:
```yaml
match_response: response.getStatusCode() == 201 & request.getMethod() == 'PUT'
```

wickedOne:
fixes #270


wickedOne:
hmz, this probably should be done agains the 1.3 branch…

PR #272, opened 2016-01-07
include request object in match_response

### LINKED ISSUE #270: include request object in match_response

wickedOne:
fixes #270

### DIFF
--- a/Tests/Unit/Http/RuleMatcherTest.php
+++ b/Tests/Unit/Http/RuleMatcherTest.php
@@ -45,4 +45,16 @@ class RuleMatcherTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($ruleMatcher->matches(new Request(), new Response('', 100)));
         $this->assertTrue($ruleMatcher->matches(new Request(), new Response('', 500)));
     }
+
+    public function testMatchResponseRequest()
+    {
+        $ruleMatcher = new RuleMatcher(new RequestMatcher(), array('match_response' => 'response.getStatusCode() == 201 & request.getMethod() == \'PUT\''));
+        $request = new Request();
+
+        $request->setMethod('POST');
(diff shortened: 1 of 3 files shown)


Tests/Unit/Http/RuleMatcherTest.php
```diff
@@ -45,4 +45,16 @@ public function testMatchResponse()
         $this->assertFalse($ruleMatcher->matches(new Request(), new Response('', 100)));
         $this->assertTrue($ruleMatcher->matches(new Request(), new Response('', 500)));
     }
+
+    public function testMatchResponseRequest()
+    {
+        $ruleMatcher = new RuleMatcher(new RequestMatcher(), array('match_response' => 'response.getStatusCode() == 201 & request.getMethod() == \'PUT\''));
```
dbu: could this not be solved with `new RuleMatcher(new RequestMatcher(null, null, 'PUT'), array('match_response' => 'response.getStatusCode() == 201))` ?

i am reluctant to add redundant possibilities to achieve the same, as that is always confusing.

wickedOne: you're absolutely right; duplicating my match rule, restricting it to a ``PUT`` method and a ``201`` response code does exactly that… 

thanks for clarifying, must have been asleep yesterday :sleeping:

ddeboer:
No problem, @wickedOne.

PR #273, opened 2016-01-08
prevent allowing cache if hash has changed

bastnic:
I heavily use role provider user context hash  with a ttl of 900s.

A session deconnexion happening while the hash is in cache trigger a situation where you set an anonymous data for an authenticated hash. In this situation there is no breach of security (yeah) but it's definitively a bug.

Proposed solution is to generate a hash at each request and prevent setting cache if hash are not equivalent (ie: roles or situation has changed).

Tests to be added, but I want to start the discussion.

### DIFF
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -131,7 +131,8 @@ class UserContextSubscriber implements EventSubscriberInterface
 
         // hash has changed, session has most certainly changed, prevent setting incorrect cache
         if ($request->attributes->has($this->hashHeader) && $request->attributes->get($this->hashHeader) !== $request->headers->get($this->hashHeader)) {
-            $response->setPrivate();
+            $response->setClientTtl(0);
+            $response->headers->addCacheControlDirective('no-cache');
 
             return;
         }
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -88,9 +88,10 @@ class UserContextSubscriber implements EventSubscriberInterface
         }
 
         $hash = $this->hashGenerator->generateHash();
-        $event->getRequest()->attributes->set($this->hashHeader, $hash);
 
         if (!$this->requestMatcher->matches($event->getRequest())) {
+            $event->getRequest()->attributes->set($this->hashHeader, $hash);
+
             return;
         }
 
@@ -129,7 +130,7 @@ class UserContextSubscriber implements EventSubscriberInterface
         $request = $event->getRequest();
 
         // hash has changed, session has most certainly changed, prevent setting incorrect cache
-        if ($request->attributes->get($this->hashHeader) !== $request->headers->get($this->hashHeader)) {
+        if ($request->attributes->has($this->hashHeader) && $request->attributes->get($this->hashHeader) !== $request->headers->get($this->hashHeader)) {
             $response->setPrivate();
 
             return;
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -49,6 +49,11 @@ class UserContextSubscriber implements EventSubscriberInterface
     /**
      * @var string
      */
+    private $hash;
+
+    /**
+     * @var string
+     */
     private $hashHeader;
 
     /**
@@ -69,6 +74,7 @@ class UserContextSubscriber implements EventSubscriberInterface
         $this->requestMatcher        = $requestMatcher;
         $this->hashGenerator         = $hashGenerator;
         $this->userIdentifierHeaders = $userIdentifierHeaders;
+        $this->hash                  = null;
         $this->hashHeader            = $hashHeader;
         $this->ttl                   = $ttl;
     }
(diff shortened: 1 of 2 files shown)


EventListener/UserContextSubscriber.php
```diff
@@ -87,11 +87,13 @@ public function onKernelRequest(GetResponseEvent $event)
             return;
         }
 
+        $hash = $this->hashGenerator->generateHash();
+        $event->getRequest()->attributes->set($this->hashHeader, $hash);
```
dbu: why store this in the request and not simply as $this->hash? or could there be issues with symfony nested requests? i would expect it to instantiate a new subscriber but not sure...

EventListener/UserContextSubscriber.php
```diff
@@ -124,6 +126,15 @@ public function onKernelResponse(FilterResponseEvent $event)
         }
 
         $response = $event->getResponse();
+        $request = $event->getRequest();
+
+        // hash has changed, session has most certainly changed, prevent setting incorrect cache
+        if ($request->attributes->get($this->hashHeader) !== $request->headers->get($this->hashHeader)) {
```
dbu: below we have a different logic: if the hashHeader is not present, we vary on the userIdentifierHeaders. i am afraid this line could break at least the hash lookup request.

looking at the code below, i wonder why we force varying on the hash. there could be parts of the site that do not depend on the user and don't want to vary on the hash... but that question is out of scope for this PR.

maybe only do the sanity check if there was a hash header in the request, in the `if` case below?

dbu:
thanks, i think its a good idea to sanity check the hash!

EventListener/UserContextSubscriber.php
```diff
@@ -124,6 +126,15 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+            $response->setPrivate();
+
```
joelwurtz: Cache should be set to "no-cache" and not private IMO

EventListener/UserContextSubscriber.php
bastnic: - agree with the break, this should be not happened if this->requestMatcher->matches. Will provide an updated version
- agree also with the too generic user context hash that can happened on resources like /translations, js, css. We prevent that in our varnish vcl with some return(hash) before the user context lookup. But i'm not sure what needs to be done on this topic.

EventListener/UserContextSubscriber.php
bastnic: No opinion on this matter, you tell.

joelwurtz: I think it's safe to have this under $this->hash, as there is no reason a hash will be changed during a sub request.

EventListener/UserContextSubscriber.php
joelwurtz: :+1:

bastnic:
- changed private to no-cache
- attribute set only on non user-context-hash request

EventListener/UserContextSubscriber.php
```diff
@@ -87,11 +87,14 @@ public function onKernelRequest(GetResponseEvent $event)
             return;
         }
 
+        $hash = $this->hashGenerator->generateHash();
+
         if (!$this->requestMatcher->matches($event->getRequest())) {
+            $event->getRequest()->attributes->set($this->hashHeader, $hash);
```
bastnic: you prefer here a $this->hash = $hash

EventListener/UserContextSubscriber.php
```diff
@@ -124,6 +127,16 @@ public function onKernelResponse(FilterResponseEvent $event)
         }
 
         $response = $event->getResponse();
+        $request = $event->getRequest();
+
+        // hash has changed, session has most certainly changed, prevent setting incorrect cache
+        if ($request->attributes->has($this->hashHeader) && $request->attributes->get($this->hashHeader) !== $request->headers->get($this->hashHeader)) {
```
bastnic: and here a $this->hash not null ?

EventListener/UserContextSubscriber.php
dbu: yes, i would say so

EventListener/UserContextSubscriber.php
```diff
@@ -124,6 +127,16 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+            $response->setClientTtl(0);
+            $response->headers->addCacheControlDirective('no-cache');
+
+            return;
+        }
+
         $vary = $response->getVary();
 
         if ($event->getRequest()->headers->has($this->hashHeader)) {
```
dbu: i would move the sanity check in here. and remove the `if` to use $request.

dbu:
the sanity check breaks some of the tests. those tests use mocking, so likely its correct to adjust the tests.

i think this pull request could go against 1.3 as its preventing a bug with showing anonymous data to logged in users (no confidential information, so no security break, but still a break). master is version 2 and that will still take a bit of time until it gets released.

bastnic:
Yep, tests needs to be fixed and more tests should be added to 1/ verify the bug 2/ verify the fix. 

And this is a production bug on a big website, and it generates very weird and stressing bugs. I'm surely not the only one to have that. So +1 for 1.3.

dbu:
agreed. can you create a branch from the 1.3 branch and cherry-pick your
commit in there and start a new pull request with that? i will then
merge the 1.3 branch into master once the fix is merged.

bastnic:
@dbu will do, once tests are green

dbu:
they look green!

though there might be differences in the 1.3 branch with regard to testing. but you will see when moving your commits to the new branch...

bastnic:
@dbu done

dbu:
thanks. closing in favor of #274

PR #274, opened 2016-01-08
check sanity of user context hash, if not, prevent setting any cache

bastnic:
I heavily use role provider user context hash  with a ttl of 900s.

A session deconnexion happening while the hash is in cache trigger a situation where you set an anonymous data for an authenticated hash. In this situation there is no breach of security (yeah) but it's definitively a bug.

Proposed solution is to generate a hash at each request and prevent setting cache if hash are not equivalent (ie: roles or situation has changed).

Tests to be added, but I want to start the discussion.

See #273

### DIFF
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -49,6 +49,11 @@ class UserContextSubscriber implements EventSubscriberInterface
     /**
      * @var string
      */
+    private $hash;
+
+    /**
+     * @var string
+     */
     private $hashHeader;
 
     /**
@@ -88,6 +93,10 @@ class UserContextSubscriber implements EventSubscriberInterface
         }
 
         if (!$this->requestMatcher->matches($event->getRequest())) {
+            if ($event->getRequest()->headers->has($this->hashHeader)) {
+                $this->hash = $this->hashGenerator->generateHash();
+            }
+
             return;
         }
 
@@ -124,9 +133,19 @@ class UserContextSubscriber implements EventSubscriberInterface
         }
 
         $response = $event->getResponse();
+        $request = $event->getRequest();
+
         $vary = $response->getVary();
 
-        if ($event->getRequest()->headers->has($this->hashHeader)) {
+        if ($request->headers->has($this->hashHeader)) {
+            // hash has changed, session has most certainly changed, prevent setting incorrect cache
+            if (!is_null($this->hash) && $this->hash !== $request->headers->get($this->hashHeader)) {
+                $response->setClientTtl(0);
+                $response->headers->addCacheControlDirective('no-cache');
+
+                return;
+            }
+
             if (!in_array($this->hashHeader, $vary)) {
                 $vary[] = $this->hashHeader;
             }
--- a/Tests/Unit/EventListener/UserContextSubscriberTest.php
+++ b/Tests/Unit/EventListener/UserContextSubscriberTest.php
@@ -104,6 +104,7 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
 
         $requestMatcher = $this->getRequestMatcher($request, false);
         $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
+        $hashGenerator->shouldReceive('generateHash')->andReturn('hash');
 
         $userContextSubscriber = new UserContextSubscriber($requestMatcher, $hashGenerator, array('X-SessionId'), 'X-Hash');
         $event = $this->getKernelRequestEvent($request);
@@ -166,6 +167,67 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
         $userContextSubscriber->onKernelResponse($event);
 
         $this->assertEquals('X-SessionId', $event->getResponse()->headers->get('Vary'));
+    }
+
+    /**
+     * If there is no hash in the request, vary on the user identifier.
+     */
+    public function testFullRequestHashOk()
+    {
+        $request = new Request();
+        $request->setMethod('GET');
+        $request->headers->set('X-Hash', 'hash');
+
+        $requestMatcher = $this->getRequestMatcher($request, false);
+        $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
+        $hashGenerator->shouldReceive('generateHash')->andReturn('hash');
+
+        // onKernelRequest
+        $userContextSubscriber = new UserContextSubscriber($requestMatcher, $hashGenerator, array('X-SessionId'), 'X-Hash');
+        $event = $this->getKernelRequestEvent($request);
+
+        $userContextSubscriber->onKernelRequest($event);
+
+        $response = $event->getResponse();
+
+        $this->assertNull($response);
+
+        // onKernelResponse
+        $event = $this->getKernelResponseEvent($request);
+        $userContextSubscriber->onKernelResponse($event);
+
+        $this->assertContains('X-Hash', $event->getResponse()->headers->get('Vary'));
+    }
+
+    /**
+     * If there is no hash in the requests but session changed, prevent setting bad cache
+     */
+    public function testFullRequestHashChanged()
+    {
+        $request = new Request();
+        $request->setMethod('GET');
+        $request->headers->set('X-Hash', 'hash');
+
+        $requestMatcher = $this->getRequestMatcher($request, false);
+        $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
+        $hashGenerator->shouldReceive('generateHash')->andReturn('hash-changed');
+
+        // onKernelRequest
+        $userContextSubscriber = new UserContextSubscriber($requestMatcher, $hashGenerator, array('X-SessionId'), 'X-Hash');
+        $event = $this->getKernelRequestEvent($request);
+
+        $userContextSubscriber->onKernelRequest($event);
+
+        $response = $event->getResponse();
+
+        $this->assertNull($response);
+
+        // onKernelResponse
+        $event = $this->getKernelResponseEvent($request);
+        $userContextSubscriber->onKernelResponse($event);
+
+        $this->assertFalse($event->getResponse()->headers->has('Vary'));
+        $this->assertEquals('max-age=0, no-cache, private', $event->getResponse()->headers->get('Cache-Control'));
     }
 
     protected function getKernelRequestEvent(Request $request, $type = HttpKernelInterface::MASTER_REQUEST)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 Changelog
 =========
 
+
+1.3.7
+-----
+
+* Add a sanity check on UserContextHash to avoid invalid content being cached
+  (example: anonymous cache set for authenticated user). This scenario occures
+  when the UserContextHash is cached by Varnish via 
+  `fos_http_cache.user_context.hash_cache_ttl` > 0 and the session is lost via 
+  garbage collector. The data given is the anonymous one despite having a hash 
+  for authenticated, all authenticated users will then have the anonymous version.
+  Same problem could occurs with users having is role changed or anything else
+  that can modify the hash.
+
 1.3.2
 -----
 


Tests/Unit/EventListener/UserContextSubscriberTest.php
```diff
@@ -168,6 +169,67 @@ public function testOnKernelResponseNotCached()
...
+
+        // onKernelResponse
+        $event = $this->getKernelResponseEvent($request);
+        $userContextSubscriber->onKernelResponse($event);
+
+        $this->assertContains('X-Hash', $event->getResponse()->headers->get('Vary'));
+    }
+
+    /**
+     * If there is no hash in the the requests but session changed, prevent setting bad cache
```
dbu: "the the"

dbu:
looks good to me. @ddeboer can you please also have a look into this? also @andrerom or @lolautruche do you see anything that could be relevant for you?

dbu:
@bastnic can you please add a line to CHANGELOG.md?

EventListener/UserContextSubscriber.php
```diff
@@ -69,6 +74,7 @@ public function __construct(
         $this->requestMatcher        = $requestMatcher;
         $this->hashGenerator         = $hashGenerator;
         $this->userIdentifierHeaders = $userIdentifierHeaders;
+        $this->hash                  = null;
```
ddeboer: This isn’t necessary, is it?

ddeboer:
Looks good to me. Is there any way we can add an integration test for this?

EventListener/UserContextSubscriber.php
dbu: its the default anyways. i think the line can be removed.

dbu:
the only partial functional test for user context i found is https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Tests/Functional/Security/Http/Logout/ContextInvalidationLogoutHandlerTest.php

a few tests in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/tree/master/Tests/Functional/EventListener would be nice. but i think the unit tests plus confirmation that this works in the project of @bastnic would be enough for me to merge.

bastnic:
I need to tests again now that i refactored a little. The unit tests added kind of confirm the integration but I will investigate more on that.

dbu:
great. ping me when you tested it.

bastnic:
@dbu thanks for the help

andrerom:
Looks ok from my pov, but as @lolautruche is off I'll ping @bdunogier for a quick second sanity.

bastnic:
Hum, quite difficult to add a full integration test.

What I do on my scenario, fill free to implement it with phpunit ;)

```
fos_http_cache:
    user_context:
        user_identifier_headers:
            - X-Userid
        role_provider: true
        hash_cache_ttl: 900
```
// scénario 1
- purge varnish so you have no previous cache
- configure 
- authenticate, hash in cache in varnish for 900s
- kill the session storage to simulate gc.
- go to a page with two ESIs : one with vary on user id one with default cache so only user context hash
- both ESI shows non authenticated content. The first one, with vary on user id has no impact whatsoever, but the one with user context hash just filled an authenticated cache with anonymous data

- open another browser or anonymous session, just log with a user that has the same roles so the same user context hash => you will never see the authenticated version of the ESI with the vary on user context hash, you are stuck with the anonymous one for the lifetime on this cache. On my website, 14400s, 4h. Yep, you can only purge varnish…

// scénario 2
I also tested with this scenario with is funnier : 
- purge varnish so you have no previous cache
- configure 
- authenticate, hash in cache in varnish for 900s
- go to a page with two ESIs : one with vary on user id one with default cache so only user context hash. In a onKernelResponse event subscriber, add a `$session->migrate(true)` the ESI will have the old session id
- both ESI shows non authenticated content. The first one, with vary on user id has no impact whatsoever, but the one with user context hash just filled an authenticated cache with anonymous data
- refresh (without the session migrate) : the first ESI shows the correct content (it uses a vary on a parameter which has changed) but the other one not (the user context hash is the same for both old and new session). 
 
// conclusion
With the patch, on the first attempt with the old session, we check the hash and get an anonymous one, which differs from the one given as parameter by varnish. So the situation changed, we prevent setting any cache. 

Maybe we could go further and try to remove the user context hash  ? The worst thing that can happened is that we use the content for authenticated content for pages that are already in cache during the duration of the user context cache hash.

bastnic:
ping @dbu

dbu:
i think this sounds like a full integration test, not simply a functional test. we seem to have an integration test base case: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Test/ProxyTestCase.php

from .travis.yml i see that we do not have any integration tests with the bundle yet. i think for all the variation of edge cases, unit tests and maybe some functional tests should be enough. but for the full scenario you describe, the ProxyTestCase sounds like a good idea. so if you want to set this up, please do. look at the .travis.yml of FOSHttpCache how to add varnish (its enough to test with varnish 4, the various proxies differences are tested in the library) and then set up http://foshttpcachebundle.readthedocs.org/en/latest/features/testing.html in our testing things. i would do that in a new namespace Tests\Integration\... rather than under Functional. the intention of this test is to see if the interaction between varnish and symfony works.

if you have the time to do this, it will be interesting and useful. but we don't absolutely need that. a few functional tests with mocking things as needed would sound enough to me. maybe just do the functional test, then we merge this fix and you work on a few integration tests in a separate PR if you want to do it.

> Maybe we could go further and try to remove the user context hash ? The worst thing that can 
> happened is that we use the content for authenticated content for pages that are already in cache 
> during the duration of the user context cache hash.

you mean removing it on the response? you would need to also vary on the user_identifier_headers, or you get a real security issue. how would that be better than simply setting the response to uncacheable?

i wonder if we could detect the situation and trigger an invalidation of the current users cached hash lookup? then the user would revert to use the correct cache.

bastnic:
@dbu i'm putting it in prod for real tests.

I just encounter a potentiel flaw : one of our ContextProvider was not session proof and was starting session. It means that on the user context provider call we were starting session all the time. Not cool but not enormous.

But now that the hash is calculated at each request, we see a set cookie even for anonymous page. 

But on our end so, but it leads me to the conclusion that hey, we are calculating a hash for nothing if a previous token is not sent with the request.

EventListener/UserContextSubscriber.php
```diff
@@ -87,11 +92,14 @@ public function onKernelRequest(GetResponseEvent $event)
...
 
+        $hash = $this->hashGenerator->generateHash();
+
         if (!$this->requestMatcher->matches($event->getRequest())) {
+            $this->hash = $hash;
+
             return;
         }
 
-        $hash = $this->hashGenerator->generateHash();
```
bastnic: something like that:

```php
        if (!$this->requestMatcher->matches($event->getRequest())) {
            if ($event->getRequest()->headers->get($this->hashHeader)) {
                $this->hash = $this->hashGenerator->generateHash();
            }

            return;
        }

        $hash = $this->hashGenerator->generateHash();
```

dbu: hm, good point. i think this idea makes sense. we don't want to trigger a session for anonymous here. and we don't use the hash in the sanity check below if there was no hash header.

bastnic:
i'm back with production tests: patch is working!
I see an increase in load, caused by these possible facts:
- I was serving bad cached version instead of a not cached version
- We add the hash compute on every request. Depending on the algorithm chosen, it can increase a lot.
- ?

But still, I prefer a working website ;).

dbu:
did you check the cache hit ratio? did you sanity check if you still get cache hits for logged in users at all? if both are true, i guess what you see is indeed the hash calculation. timed out sessions should probably not be all that frequent that the have a visible impact, no?

you could analyze the execution of one request with xdebug and cachegrind or maybe blackfire.io to see if the hash makes such a visible part of the request. as the hash is solely between varnish and the symfony application, it needs not be an elaborate/secure thing, as long as its unique.

would you say this is ready for merge? i'd also rather have a load impact than the type of bug we have without this fix.

dbu:
@bastnic are you using your fix in production? should i merge this here?

bastnic:
I was waiting for real data, and we can safely assume that it's ok now: never have an issue related to that topic again. 

It was a hard to catch kind of a bug and I will be happy to see it merged.

bastnic:
ping @joelwurtz for confirmation

lolautruche:
+1 code wise

dbu:
thanks for this fix!

dbu:
released as 1.3.7 and porting to master now.

andrerom:
Just for reference @bdunogier is investigating [regression](https://jira.ez.no/browse/EZP-25505) for eZ Publish / eZ Platform.

So if anyone else has issues as well, please share :)

dbu:
ups. did we miss a case?

i will be on holiday starting from tomorrow - i hope @ddeboer can handle review and merge and release a patch 1.3.x version.

joelwurtz:
I don't think this is a bug, as maybe the generated hash from the exemple does not correspond to the given hash.

However in the response there is both no-cache and public, we should maybe clear the Cache-Control header WDYT ?

EventListener/UserContextSubscriber.php
```diff
@@ -124,9 +133,19 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+        $request = $event->getRequest();
+
         $vary = $response->getVary();
 
-        if ($event->getRequest()->headers->has($this->hashHeader)) {
+        if ($request->headers->has($this->hashHeader)) {
+            // hash has changed, session has most certainly changed, prevent setting incorrect cache
+            if (!is_null($this->hash) && $this->hash !== $request->headers->get($this->hashHeader)) {
+                $response->setClientTtl(0);
+                $response->headers->addCacheControlDirective('no-cache');
```
dbu: indeed we might need to be more clear with these, override the cache control directives for private no-cache and s-maxage=0 and max-age=0.

also if we use the special custom ttl header we should set it to 0 and make sure its not added again by the header listener.

dbu:
"Cache-Control: max-age=1, no-cache, public, s-maxage=300" looks weird. why max-age=1, that is not from us? but yes, commented where we should change the listener of this PR. also, we should see about priorities, this listener should come *after* CacheControlSubscriber to make sure it overrides whatever rules where set.

the topic of anonymous header is also interesting. i would recommend for ezpublish to do the hash lookup for anonymous. it will cost one request per cache lifetime of the hash lookup, which is not much. otherwise you would need to calculate the hash for anonymous users and if that ever changes (e.g. change of roles for anonymous users) caching would stop for anonymous. i would prefer that over adding a magic anonymous user header that this listener would detect in the sanity check.

andrerom:
be aware in our case it looks like hash indeed changed, but I'll let @bdunogier comment on that for what investigation revealed so far.

bdunogier:
Not related to this patch, afaict. The problem is on our side (our siteaccess / context aware config excludes this route, meaning that no user is logged in in the _fos_user sub-request. Kind of a major problem, but from a different source :-)

dbu:
i fear that we might started hiding this sort of problem when we
switched from the security context to the token storage, as token
storage is always available afaik:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/UserContext/RoleProvider.php#L61-L67

if you have an idea how we can harden this code to prevent such a
mistake, please tell.

andrerom:
@lolautruche any input on the overall design here from you're pov?

lolautruche:
After giving a second look, I think there is something wrong. Hash generation will indeed be triggered on **every requests**, except hash lookup. The first drawback I can see is potential performance impact, which can be important on eZ, due to the way we generate the hash.
Furthermore, there when a user is anonymous and has no session, there is no hash lookup to improve performance, and a hardcoded hash is used. This hash will always be different from the one generated on each request. This is more than likely the problem described here: https://jira.ez.no/browse/EZP-25505

IMO, instead of comparing hashes, we should compare session ID **if and only if** session is started (note that session cookie is passed to internal hash lookup request).

dbu:
we discussed the performance impact and i [stated](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/274#issuecomment-171894367) that i prefer a performance impact (on cache misses) over delivering mixed up caches when a session timed out. is the ez cache calculation that expensive? can you maybe find something to make the lookup of the hash more efficient? (without running into problems when permissions of a user change)

i don't think the session is good enough. we could be using basic authentication. and either way, somebody could change the permissions of a logged in user, resulting in the request being done with different permissions than where used to calculate the hash. which would lead to mixed up caches.

andrerom:
We could consider speeding this up by moving it further down in our layers to be able to cache it, or we can add cache for this higher up but then we will potentially run into layering/invalidation issues if not done right.

Right now we cache the entities involved, but not the calculations which can potentially have to iterate over quite a bit og roles and policies for it on a complex setup.

bdunogier:
One question about the `anonymous_hash`.

We're having problems because our anonymous hash varies depending on how the anonymous role is configured. This wasn't a problem before this PR, since we never tried to generate a hash for anonymous users.

Now we have [this call](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/274/files#diff-5e8d072188e13a501d7eef85ec20ca6cR97) that does generate it, and of course, it doesn't match the hardcoded one that was used by the `UserContextSubscriber`.

I'm thinking we should also use the `anonymous_hash` in that case. Making it configurable would force us to also configure it in the VCL, while I do not think it is necessary (among other consequences).

andrerom:
So `$this->hashGenerator->generateHash()` does not take the hardcoded hash into account? That explains it.

lolautruche:
> So $this->hashGenerator->generateHash() does not take the hardcoded hash into account?

Indeed not, as goal of hard-coded anonymous hash was to avoid to boot the kernel.

andrerom:
true

bdunogier:
I'm at the bank right now (yay), but I have a properly designed patch in mind for this. I'll send a PR before the end of the day.

dbu:
okay, i guess this should not have gone into 1.3 but rather 1.4 or 2... i considered it a security relevant fix - but the side effects where unexpected. glad you found a solution @bdunogier, just commented on #289

PR #275, opened 2016-01-14
Support for caching OPTIONS HTTP verb

renatomefidf:
Hello,

I'm was trying to cache an OPTIONS request, it was not registering any kind of header, while debugging I found this:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/CacheControlSubscriber.php#L105

Which leads to:
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L1473

Meaning I can't handle OPTIONS requests.

My approach was to make it safe during the Subscriber event. Reason being, we do not want to change the safe methods on SF core.

I'm not using new array notation since I saw the code was using the old one!

Is this a good way? Please I would like to receive any feedback!

### DIFF
--- a/EventListener/CacheControlSubscriber.php
+++ b/EventListener/CacheControlSubscriber.php
@@ -101,8 +101,15 @@ class CacheControlSubscriber extends AbstractRuleSubscriber implements EventSubs
             $response->headers->set($this->debugHeader, 1, false);
         }
 
+
+        if ($this->skip) {
+            return;
+        }
+
         // do not change cache directives on unsafe requests.
-        if ($this->skip || !$this->isRequestSafe($request)) {
+        if ($request->getMethod() !== 'OPTIONS' &&
+            !$this->isRequestSafe($request)
+        ) {
             return;
         }
 
--- a/Tests/Unit/EventListener/CacheControlSubscriberTest.php
+++ b/Tests/Unit/EventListener/CacheControlSubscriberTest.php
@@ -353,6 +353,38 @@ class CacheControlSubscriberTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * Safe methods should get into the matchRule
+     * @dataProvider getSafeMethods
+     * @param $method
+     */
+    public function testSafeMethodsTriggersMatchRule($method)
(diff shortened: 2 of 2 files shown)


EventListener/CacheControlSubscriber.php
```diff
@@ -101,8 +101,15 @@ public function onKernelResponse(FilterResponseEvent $event)
...
         }
 
+
+        if ($this->skip) {
+            return;
+        }
+
         // do not change cache directives on unsafe requests.
-        if ($this->skip || !$this->isRequestSafe($request)) {
+        if ($request->getMethod() !== 'OPTIONS' &&
```
renatomefidf: Not using Request::METHOD_OPTIONS since it was failing the tests on travis

renatomefidf:
Now checking RFC I saw:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

```
9.2 OPTIONS

The OPTIONS method represents a request for information about the communication options available on the request/response chain identified by the Request-URI. This method allows the client to determine the options and/or requirements associated with a resource, or the capabilities of a server, without implying a resource action or initiating a resource retrieval.

Responses to this method are not cacheable.
```

So, should close it?

dbu:
i would rather not add this here. 

but the CacheControlSubscriber::isRequestSafe method is protected on purpose - you can extend the event listener and overwrite this method if in your application you know that options can be cached.

that said, not sure why they are considered not cacheable - i would assume this usually depends on permissions of the user. so together with the user context, it could make sense to cache them if you have a substantial load and the OPTIONS requests are slow. but as said, extend the listener rather than we change this class to non-standard HTTP behaviour.

dbu:
but thanks for wanting to help. don't let this deter you from sending further pull requests and discuss things, we appreciate all input.

renatomefidf:
@dbu thanks for the feedback!

I also couldn't find the exactly reason why OPTIONS is non cacheable, this seems like a historic choice!

Indeed, if I find something else I'll look into making a new contribution!

Have a nice weekend!

PR #280, opened 2016-02-09
flashbag message doc precision

nicolasmure:
Hi,

This is a little precision in the flash messages documentation.
I had some troubles when I was using the current example (the flash message was still appearing after page reload), but this tweak solved it.
I hope it could help :smile:


dbu:
good point, thanks!

PR #282, opened 2016-02-17
trigger deprecation notices for deprecated code

### LINKED ISSUE #266: Trigger deprecations with @trigger_error
Methods in the CacheManager have been deprecated. But there are no deprecations triggered.
So I'd suggest to add ``@trigger_error` calls as symfony does.

dbu:
fix #266

@Tobion am i correct that this should go into the 1.3 branch? we want to delete this things for 2.0 so the deprecations would not really help there. i don't want to build a 1.4 version only for the migration like symfony 2.8 is - too much effort.
also, do you know how to make the deprecations helper not complain about our legacy tests? i tried the `@legacy` annotation on the test cases, but that did not seem help.

### DIFF
--- a/HttpCache.php
+++ b/HttpCache.php
@@ -20,6 +20,8 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
+@trigger_error('The '.__NAMESPACE__.'\HttpCache class is deprecated since version 1.2 and will be removed in 2.0. Use FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache instead.', E_USER_DEPRECATED);
+
 /**
  * Base class for enhanced Symfony reverse proxy based on the symfony FrameworkBundle HttpCache.
  *
--- a/CacheManager.php
+++ b/CacheManager.php
@@ -72,6 +72,8 @@ class CacheManager extends CacheInvalidator
      */
     public function tagResponse(Response $response, array $tags, $replace = false)
     {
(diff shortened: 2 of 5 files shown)


dbu:
ah, found the thing about legacy tests. `@group legacy`, not `@legacy`...

HttpCache.php
```diff
@@ -83,6 +83,8 @@
 
     public function __construct(HttpKernelInterface $kernel, $cacheDir = null)
     {
+        @trigger_error('The '.__CLASS__.' class is deprecated since version 1.2 and will be removed in 2.0. Use FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache instead.', E_USER_DEPRECATED);
```
Tobion: When the whole class is deprecated, you could also put this above the class definition directly.

Tobion:
If there won't be a 1.4 version (the next one is 2.0), then it should indeed make it into 1.3

HttpCache.php
```diff
@@ -20,6 +20,8 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
+@trigger_error('The '.__CLASS__.' class is deprecated since version 1.2 and will be removed in 2.0. Use FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache instead.', E_USER_DEPRECATED);
```
dbu: @Tobion like this?

Tobion: __CLASS__ is not defined outside of a class. Use something like this: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/Extension/Core/ChoiceList/ObjectChoiceList.php#L14

dbu: ups. thanks

Issue #277, opened 2016-01-29
Issue with FOSRestBundle


nicolasmure:
Hi,

Thank you for this bundle!
I'm building an app with an admin panel (html) and a REST API (json). I have FOSRestBundle installed. I wanted to add FOSHttpCacheBundle to control Varnish 3.0.7 cache so I followed [the doc](http://foshttpcachebundle.readthedocs.org/en/latest/index.html) : [varnish config](http://foshttpcache.readthedocs.org/en/stable/varnish-configuration.html), [caching headers](http://foshttpcachebundle.readthedocs.org/en/latest/features/headers.html), [user context](http://foshttpcachebundle.readthedocs.org/en/latest/features/user-context.html).

Here is my config for `fos_http_cache` :
```yaml
fos_http_cache:
    proxy_client:
        varnish:
            servers: %myapp_trusted_proxies%
            base_url: myapp.com
    user_context:
        enabled: true
        hash_cache_ttl: 600
    cache_control:
        defaults:
            overwrite: true
        rules:
            # match everything to set defaults
            -
                match:
                    path: ^/
                headers:
                    overwrite: false
                    cache_control: { public: true, max_age: 600, s_maxage: 600 }
                    etag: true

```
When I try to access `myapp.com` with a browser, I get the following error :
`406 Not Acceptable - No matching accepted Response format could be determined` thrown from `FOS\RestBundle\EventListener\FormatListener l.67`.
I have this error since I configured my varnish 3 to support user_context ([as explained in the doc](http://foshttpcache.readthedocs.org/en/stable/varnish-configuration.html#user-context)).
Apparently, the reverse-proxy is sending a request with a accept header to `application/vnd.fos.user-context-hash`, which is disturbing FOSRestBundle who's trying to determine the response format type from that header.
When I change my `fos_rest` config as it :
```yaml
fos_rest:
    param_fetcher_listener: true
    body_listener: true
    body_converter:
        enabled: true
    view:
        view_response_listener: 'force'
        # When I add these 2 lines :
        mime_types:
            html: ['application/vnd.fos.user-context-hash']

    format_listener:
        rules:
            - { path: '^/api', priorities: ['json'], fallback_format: json, prefer_extension: false }
            - { path: '^/', priorities: ['html', 'text/css'], fallback_format: text/plain }
```
[...]

PR #284, opened 2016-02-23
Added always_vary_on_context_hash option to allow disabling Vary header modification

supasteev0:
As suggested in issue #277 of FosHttpCache repository (https://github.com/FriendsOfSymfony/FOSHttpCache/issues/277), I added a configuration option called **always_vary_on_context_hash** under **user_context** configuration key.
Example : 
fos_http_cache:
    user_context:
        always_vary_on_context_hash: false

Option value is set to true by default.
If set to false, UserContextSubscriber listener will not change Vary header, so that cache will not vary even if a user is logged in .
This will allow developers to choose in their controllers whether to add the user context hash Vary header or not.


DependencyInjection/Configuration.php
```diff
@@ -611,6 +611,10 @@ private function addUserContextListenerSection(ArrayNodeDefinition $rootNode)
                             ->defaultValue(0)
                             ->info('Cache the response for the hash for the specified number of seconds. Setting this to 0 will not cache those responses at all.')
                         ->end()
+                        ->scalarNode('always_vary_on_context_hash')
+                        ->defaultValue(1)
```
dbu: please make this a booleanNode with defaultTrue

EventListener/UserContextSubscriber.php
```diff
@@ -56,12 +56,18 @@ class UserContextSubscriber implements EventSubscriberInterface
...
+    private $alwaysVary;
+
     public function __construct(
         RequestMatcherInterface $requestMatcher,
         HashGenerator $hashGenerator,
         array $userIdentifierHeaders = array('Cookie', 'Authorization'),
         $hashHeader = "X-User-Context-Hash",
-        $ttl = 0
+        $ttl = 0,
+        $alwaysVary = true
```
dbu: i think here we should call this `addVaryOnHash`

EventListener/UserContextSubscriber.php
```diff
@@ -123,22 +130,25 @@ public function onKernelResponse(FilterResponseEvent $event)
             return;
         }
 
-        $response = $event->getResponse();
-        $vary = $response->getVary();
+        if($this->alwaysVary)
+        {
```
dbu: code style: the { should go on the same line as the if, with spaces like this: `if ($this->addVaryOnHash) {`

dbu:
cool! looks good.

you will need to adjust the configuration tests for the new flag, see https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/111199910

can you please add documentation to the configuration reference?
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Resources/doc/reference/configuration/user-context.rst - including the information that if this is not always added, the developer must make sure to set the Vary headers correctly by other means.

supasteev0:
Hello David, 

I made the changes you requested. I had some trouble getting the tests to work but it seems ok now. I also updated the configuration reference documentation. Tell me if everything is ok for you.

EventListener/UserContextSubscriber.php
```diff
@@ -71,6 +77,7 @@ public function __construct(
         $this->userIdentifierHeaders = $userIdentifierHeaders;
         $this->hashHeader            = $hashHeader;
         $this->ttl                   = $ttl;
+        $this->addVaryOnHash            = $addVaryOnHash;
```
dbu: alignment is off

Resources/doc/reference/configuration/user-context.rst
```diff
@@ -120,6 +120,16 @@ request. However, when you decide to cache hash responses, you must invalidate
...
 This bundle provides a logout handler that takes care of this for you.
 
+.. _always_vary_on_context_hash:
+
+``always_vary_on_context_hash``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``true``
+
+Set ``always_vary_on_context_hash`` to  ``false`` to disable always adding X-User-Context-Hash or Cookie, Authorization in Vary header.
```
dbu: what do you mean with "or Cookie, Authorization"? we keep setting the Vary for the hash lookup even if this is set to false - which is the correct behaviour.

Resources/doc/reference/configuration/user-context.rst
```diff
@@ -120,6 +120,16 @@ request. However, when you decide to cache hash responses, you must invalidate
...
+If this is set to false, you must make sure to manually set the Vary header correctly when the cache needs to vary on user context.
```
dbu: ... when the content of the response depends on the context of the user.

Resources/doc/reference/configuration/user-context.rst
```diff
@@ -120,6 +120,16 @@ request. However, when you decide to cache hash responses, you must invalidate
 them when the user context changes, particularly when the user logs in or out.
 This bundle provides a logout handler that takes care of this for you.
 
+.. _always_vary_on_context_hash:
```
dbu: you only need this when some documentation does a `:ref:` link to this documentation. you can remove this line, as there is no such link.

dbu:
looks good, i just commented a few tiny details. can you update those and then squash your commits into one please?

Resources/doc/reference/configuration/user-context.rst
supasteev0: Actually that is not true anymore because, if you look at the event listener changes I made, when the option is set to false the Vary header is not modified in any case (i.e, whether the hash is present in the request or not). So the Vary Cookie, Auhtorization which is usually set by this bundle  will not be set anymore.

supasteev0:
What do you mean by "squash your commits into one" ? Do I need to create a new branch with only one commit for all modifications ? And make a new pull request ?

Resources/doc/reference/configuration/user-context.rst
dbu: oh, ok. does that make sense? i think we should change the code then so that it still adds the vary for the lookup request. or do you see a use case where you would not want to vary that?

dbu:
squash is explained [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) - its an operation on the branch to make all commits into one. but don't worry, i can also squash them for you once the branch is ready to merge. 
lets first sort out the option question though.

Resources/doc/reference/configuration/user-context.rst
supasteev0: Well with an ezPublish web app and "ez customized" Varnish vcl, there is always a User Hash in the request. So the listener will only add Vary Cookie, Authorization when going through Apache directly. 
But in a more general sense, I think adding a Vary Cookie to cache different responses in Varnish depending on cookie header value (which will ALWAYS be different for any website user, especially if you set a session cookie) is a not a very good practice and can be dangerous regarding Varnish cache size which could be very quickly overloaded. That's why I decided to completely skip the Vary modification with this option. 
And for those who want to keep the Vary Cookie, Authorization it will not be a problem as it will remain the default behavior when using the http cache bundle.

dbu: but even ez does the hash lookup initially, no? i am talking about the answer to the hash lookup request - that one has to vary on the cookie, it would make no sense otherwise, no? 
for the actual content, it does not always need to vary on the hash, so there the option makes sense. @lolautruche can you maybe help us here, am i missing something?

EventListener/UserContextSubscriber.php
stof: @dbu whihc is exactly why we have decided not to use alignment in Symfony :smile:

dbu: i agree with you stof

Resources/doc/reference/configuration/user-context.rst
supasteev0: You are totally right, we need to Vary on cookie, but only for the response to the hash lookup request, and not for all responses.
It seems to me that this Vary is already set in the kernelRequest with the following code : 
`$response->setVary($this->userIdentifierHeaders);`
so there should be no need to set it again in kernelReponse.
If @lolautruche could indeed confirm that it would be great !

EventListener/UserContextSubscriber.php
```diff
@@ -123,22 +130,24 @@ public function onKernelResponse(FilterResponseEvent $event)
...
-            }
-        } else {
-            foreach ($this->userIdentifierHeaders as $header) {
-                if (!in_array($header, $vary)) {
-                    $vary[] = $header;
+            if ($event->getRequest()->headers->has($this->hashHeader)) {
+                if (!in_array($this->hashHeader, $vary)) {
+                    $vary[] = $this->hashHeader;
+                }
+            } else {
```
dbu: can you add a comment here that says something like 
```
/*
 * Additional safety: If for some reason we get requests without a user hash, vary on the credentials to avoid the caching proxy mixing up caches between users. For the hash lookup request, the headers are already added in onKernelRequest.
 */
```

Resources/doc/reference/configuration/user-context.rst
dbu: ah, i think i got it now. for the hash lookup we handle that in onKernelRequest. but onKernelResponse we still add the vary in case the hash header was not present, as a safety precaution against mistakes. the no vary header switch turns of the safety, but not the hash lookup vary.

okay, i agree. can you reformulate this to be more clear? it does not automatically add the vary on X-User-Hash and switches of the safety precaution of adding vary on the user identifier headers if no hash was in the request.

supasteev0:
Hello David, I added the comments in the event listener & added more explanations in the documentation. Then I squashed the commits. Tell me if this seems ok for you now.

Resources/doc/reference/configuration/user-context.rst
lolautruche: I'm a bit late on this one, but yeah, it seems we're safe.

lolautruche:
+1

That's indeed a nice enhancement :smiley:

dbu:
i am wrapping this up in #288 - had to solve a merge conflict and then added a changelog

supasteev0:
Hello David. Looks like this has only been merged on master for now. Do you know if it will be merged in 1.3.7 (or in a new tagged version) soon ? Just wondering how we could update the bundle in our existing projects to use the new option.

dbu:
i just came back from a long holidays, thus the long silence...
i would rather not backport this to 1.3 as its a new feature. i hope we can release 2.0 soon. 

you could use 2.0@dev for now. but you would need to update your application as there are a couple of changes. and there will be some more breaks before 2.0 can be considered stable. if you go that way, you could help us figure out if things are working already or not.

PR #288, opened 2016-02-26
Added always_vary_on_context_hash option

dbu:
continue #284


EventListener/UserContextSubscriber.php
```diff
@@ -146,18 +153,24 @@ public function onKernelResponse(FilterResponseEvent $event)
                 return;
```
dbu: @supasteev0 we added this sanity check about the hash header in a different branch, while you did your pull request. i decided even if we don't auto-set the vary header, we still do the sanity check (the typical situation would be that a session timed but the hash is still cached)

do you agree it makes sense to then still not cache the response? i check the new flag you add only after this sanity check.

supasteev0: @dbu OK, that's fine by me, and the sanity check looks like a very nice improvement !

ddeboer:
Awesome guys, nice addition! :sunglasses:

PR #289, opened 2016-03-03
Do not sanity check hash of anonymous requests

bdunogier:
> Fixes an issue described in #274

The hardcoded anonymous hash might differ from the one used in the sanity check introduced by this PR. This causes cache expirations for us since anonymous requests don't match.

This change excludes anonymous requests, using an `AnonymousRequestMatcher`, from this sanity check.

https://github.com/ezsystems/ezpublish-kernel/pull/1601 describes the consequences and way to reproduce on ezplatform.

## TODO
- [x] Update tests
- [x] Check inconsistency with what is used for the Vary header

### DIFF
--- a/EventListener/UserContextSubscriber.php
+++ b/EventListener/UserContextSubscriber.php
@@ -13,6 +13,7 @@ namespace FOS\HttpCacheBundle\EventListener;
 
 use FOS\HttpCache\UserContext\HashGenerator;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestMatcherInterface;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
@@ -61,12 +62,22 @@ class UserContextSubscriber implements EventSubscriberInterface
      */
     private $ttl;
 
+    /**
+     * Used to exclude anonymous requests (no authentication nor session) from user hash sanity check.
+     * It prevents issues when the hash generator that is used returns a customized value for anonymous users,
+     * that differs from the documented, hardcoded one.
+     *
+     * @var RequestMatcherInterface
+     */
+    private $anonymousRequestMatcher;
+
     public function __construct(
         RequestMatcherInterface $requestMatcher,
         HashGenerator $hashGenerator,
         array $userIdentifierHeaders = array('Cookie', 'Authorization'),
         $hashHeader = 'X-User-Context-Hash',
-        $ttl = 0
+        $ttl = 0,
+        RequestMatcherInterface $anonymousRequestMatcher = null
     ) {
         if (!count($userIdentifierHeaders)) {
             throw new \InvalidArgumentException('The user context must vary on some request headers');
@@ -76,6 +87,7 @@ class UserContextSubscriber implements EventSubscriberInterface
         $this->userIdentifierHeaders = $userIdentifierHeaders;
         $this->hashHeader = $hashHeader;
         $this->ttl = $ttl;
+        $this->anonymousRequestMatcher = $anonymousRequestMatcher;
     }
 
     /**
@@ -93,7 +105,7 @@ class UserContextSubscriber implements EventSubscriberInterface
         }
 
         if (!$this->requestMatcher->matches($event->getRequest())) {
-            if ($event->getRequest()->headers->has($this->hashHeader)) {
+            if ($event->getRequest()->headers->has($this->hashHeader) && !$this->isAnonymous($event->getRequest())) {
                 $this->hash = $this->hashGenerator->generateHash();
             }
 
@@ -118,6 +130,20 @@ class UserContextSubscriber implements EventSubscriberInterface
         }
 
         $event->setResponse($response);
+    }
+
+    /**
+     * Tests if $request is an anonymous request or not.
+     *
+     * For backward compatibility reasons, true will be returned if no anonymous request matcher was provided.
+     *
+     * @param Request $request
+     *
+     * @return bool
+     */
+    private function isAnonymous(Request $request)
+    {
+        return $this->anonymousRequestMatcher ? $this->anonymousRequestMatcher->matches($request) : false;
     }
 
     /**
--- a/Resources/config/user_context.xml
+++ b/Resources/config/user_context.xml
@@ -27,6 +27,7 @@
             <argument />
             <argument />
             <argument />
+            <argument type="service" id="fos_http_cache.user_context.anonymous_request_matcher" />
             <tag name="kernel.event_subscriber" />
         </service>
 
@@ -38,6 +39,10 @@
             <argument type="service" id="fos_http_cache.default_proxy_client" />
             <argument />
             <argument />
+        </service>
+
+        <service id="fos_http_cache.user_context.anonymous_request_matcher" class="FOS\HttpCacheBundle\UserContext\AnonymousRequestMatcher">
+            <argument type="collection" />
         </service>
     </services>
 </container>
--- a/UserContext/AnonymousRequestMatcher.php
+++ b/UserContext/AnonymousRequestMatcher.php
@@ -36,6 +36,11 @@ class AnonymousRequestMatcher implements RequestMatcherInterface
                 return false;
             }
             if ($request->headers->has($header)) {
+                if (strtolower($header) === 'cookie' && 0 === $request->cookies->count()) {
+                    // ignore empty cookie header
+                    continue;
+                }
+
                 return false;
             }
         }
--- a/UserContext/AnonymousRequestMatcher.php
+++ b/UserContext/AnonymousRequestMatcher.php
@@ -32,9 +32,6 @@ class AnonymousRequestMatcher implements RequestMatcherInterface
     public function matches(Request $request)
     {
         foreach ($this->userIdentifierHeaders as $header) {
-            if (strtolower($header) === 'cookie' && $request->cookies->count()) {
-                return false;
-            }
             if ($request->headers->has($header)) {
                 if (strtolower($header) === 'cookie' && 0 === $request->cookies->count()) {
                     // ignore empty cookie header
--- a/UserContext/AnonymousRequestMatcher.php
+++ b/UserContext/AnonymousRequestMatcher.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\UserContext;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
+/**
+ * Matches anonymous requests using a list of identification headers.
+ */
+class AnonymousRequestMatcher implements RequestMatcherInterface
+{
+    private $userIdentifierHeaders;
+
+    /**
+     * @param array $userIdentifierHeaders List of request headers that authenticate a non-anonymous request
+     */
+    public function __construct(array $userIdentifierHeaders)
+    {
+        $this->userIdentifierHeaders = $userIdentifierHeaders;
+    }
+
+    public function matches(Request $request)
+    {
+        foreach ($this->userIdentifierHeaders as $header) {
+            if (strtolower($header) === 'cookie' && $request->cookies->count()) {
+                return false;
+            }
+            if ($request->headers->has($header)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -192,6 +192,9 @@ class FOSHttpCacheExtension extends Extension
             ->replaceArgument(3, $config['user_hash_header'])
             ->replaceArgument(4, $config['hash_cache_ttl']);
 
+        $container->getDefinition($this->getAlias().'.user_context.anonymous_request_matcher')
+            ->replaceArgument(0, $config['user_identifier_headers']);
+
         if ($config['logout_handler']['enabled']) {
             $container->getDefinition($this->getAlias().'.user_context.logout_handler')
                 ->replaceArgument(1, $config['user_identifier_headers'])
--- a/Tests/Unit/EventListener/UserContextSubscriberTest.php
+++ b/Tests/Unit/EventListener/UserContextSubscriberTest.php
@@ -199,6 +199,39 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * If the request is an anonymous one, no hash should be generated/validated.
+     */
+    public function testFullAnonymousRequestHashNotGenerated()
+    {
+        $request = new Request();
+        $request->setMethod('GET');
+        $request->headers->set('X-Hash', 'anonymous-hash');
+
+        $requestMatcher = $this->getRequestMatcher($request, false);
+        $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
+        $hashGenerator->shouldReceive('generateHash')->never();
+
+        $anonymousRequestMatcher = \Mockery::mock('\Symfony\Component\HttpFoundation\RequestMatcherInterface');
+        $anonymousRequestMatcher->shouldReceive('matches')->andReturn(true);
+
+        // onKernelRequest
+        $userContextSubscriber = new UserContextSubscriber($requestMatcher, $hashGenerator, array('X-SessionId'), 'X-Hash', 0, $anonymousRequestMatcher);
+        $event = $this->getKernelRequestEvent($request);
+
+        $userContextSubscriber->onKernelRequest($event);
+
+        $response = $event->getResponse();
+
+        $this->assertNull($response);
+
+        // onKernelResponse
+        $event = $this->getKernelResponseEvent($request);
+        $userContextSubscriber->onKernelResponse($event);
+
+        $this->assertContains('X-Hash', $event->getResponse()->headers->get('Vary'));
+    }
+
+    /**
      * If there is no hash in the requests but session changed, prevent setting bad cache.
      */
     public function testFullRequestHashChanged()
--- a/Tests/Unit/UserContext/AnonymousRequestMatcherTest.php
+++ b/Tests/Unit/UserContext/AnonymousRequestMatcherTest.php
@@ -29,12 +29,22 @@ class AnonymousRequestMatcherTest extends PHPUnit_Framework_TestCase
     public function testNoMatchIfCookie()
     {
         $request = new Request();
-        $request->headers->set('Cookie', '');
+        $request->headers->set('Cookie', 'PHPSESSID7e476fc9f29f69d2ad6f11dbcd663b42=25f6d9c5a843e3c948cd26902385a527');
         $request->cookies->set('PHPSESSID7e476fc9f29f69d2ad6f11dbcd663b42', '25f6d9c5a843e3c948cd26902385a527');
 
         $requestMatcher = new AnonymousRequestMatcher(['Cookie', 'Authorization']);
 
         $this->assertFalse($requestMatcher->matches($request));
+    }
+
+    public function testNoMatchIfEmptyCookieHeader()
+    {
+        $request = new Request();
+        $request->headers->set('Cookie', '');
+
+        $requestMatcher = new AnonymousRequestMatcher(['Cookie', 'Authorization']);
+
+        $this->assertTrue($requestMatcher->matches($request));
     }
 
     public function testNoMatchIfAuthenticationHeader()
--- a/Tests/Unit/UserContext/AnonymousRequestMatcherTest.php
+++ b/Tests/Unit/UserContext/AnonymousRequestMatcherTest.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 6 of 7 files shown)


bdunogier:
Travis looks _very_ broken, but I doubt that my changes did that... how stable is the 1.3 branch ? :-)

ddeboer:
The test failures have to do with your introduction of `AnonymousRequestMatcher`. Just to be sure, I restarted the tests on the 1.3 branch and they are still green.

bdunogier:
Okay, thank you @ddeboer.

EventListener/UserContextSubscriber.php
```diff
@@ -121,6 +129,20 @@ public function onKernelRequest(GetResponseEvent $event)
     }
 
     /**
+     * Tests if $request is an anonymous request or not.
+     *
+     * For backward compatibility reasons, true will be returned if no anonymous request matcher was provided.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
```
dbu: please only use Request here, `use` statements work for annotations too

EventListener/UserContextSubscriber.php
```diff
@@ -61,21 +62,28 @@ class UserContextSubscriber implements EventSubscriberInterface
      */
     private $ttl;
 
+    /**
+     * @var \Symfony\Component\HttpFoundation\RequestMatcherInterface
```
dbu: please only use RequestMatcherInterface here, `use` statements work for annotations too

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,62 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\UserContext;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class AnonymousRequestMatcher implements RequestMatcherInterface
```
dbu: please add a class level comment to explain that this matcher matches all requests that are anonymous.

EventListener/UserContextSubscriber.php
```diff
@@ -61,21 +62,28 @@ class UserContextSubscriber implements EventSubscriberInterface
      */
     private $ttl;
 
+    /**
```
dbu: and please explain here that not booting the symfony kernel is the reason why we use this matcher rather than just looking at symfony and see if there is a non-anonymous user.

Resources/config/user_context.xml
```diff
@@ -10,6 +10,7 @@
         <parameter key="fos_http_cache.user_context.role_provider.class">FOS\HttpCacheBundle\UserContext\RoleProvider</parameter>
         <parameter key="fos_http_cache.user_context.request_matcher.class">FOS\HttpCacheBundle\UserContext\RequestMatcher</parameter>
         <parameter key="fos_http_cache.user_context.logout_handler.class">FOS\HttpCacheBundle\Security\Http\Logout\ContextInvalidationLogoutHandler</parameter>
+        <parameter key="fos_http_cache.user_context.anonymous_request_matcher.class">FOS\HttpCacheBundle\UserContext\AnonymousRequestMatcher</parameter>
```
dbu: parameters for class names is not best practice anymore. lets just put the class name into the service definition directly. we will do that for everything, if not already done with 2.0

dbu:
i agree with the solution. if i think this through correctly, things should work unless you manage to re-use a non-anonymous hash with an anonymous request - but that would be a misconfiguration of your application. #274 was about expired sessions, where the server side session has expired, but the session id is still sent in the request. in that case, we find the hash in the cache and do the request with hash and session id, and thus compare hashes (and see that they mismatch, or if the hash was not in the cache, see the "real" anonymous hash, vs the fake hash ez would create on a request without session cookie). @bastnic maybe you can give us your thoughts as well?

(btw, i still feel ezpublish would be better off simply do a hash lookup for anonymous users as well and then cache that - its a low overhead keeping many things simpler. for example, the anonymous cache will be duplicated between completely anonymous users and users that are determined by symfony to be anonymous because the session expired or something.)

the tests seem to say that the configuration for the anonymous matcher is wrong.

EventListener/UserContextSubscriber.php
dbu: actually, its not only performance. if we would boot the kernel, we would treat the expired session as anonymous which is exactly what #274 wanted to prevent. maybe explain the thing about anonymous fake hashes in the phpdoc here so we still know why this is done like this.

EventListener/UserContextSubscriber.php
bdunogier: Will do.

Habit we have for those using less fancy editors.

EventListener/UserContextSubscriber.php
bdunogier: Added a comment.

bdunogier:
Review remarks taken care of.

EventListener/UserContextSubscriber.php
```diff
@@ -121,6 +133,20 @@ public function onKernelRequest(GetResponseEvent $event)
...
+     *
+     * For backward compatibility reasons, true will be returned if no anonymous request matcher was provided.
+     *
+     * @param Request $request
+     *
+     * @return bool
+     */
+    private function isAnonymous(Request $request)
+    {
+        return isset($this->anonymousRequestMatcher) ? $this->anonymousRequestMatcher->matches($request) : false;
```
dbu: lets remove the isset here and just say $this->anonymousRequestMatcher ? $this-... or if you don't like this null !== $this->anonymousRequestMatcher ?

Resources/config/user_context.xml
```diff
@@ -39,5 +40,9 @@
             <argument />
             <argument />
         </service>
+
+        <service id="fos_http_cache.user_context.anonymous_request_matcher" class="FOS\HttpCacheBundle\UserContext\AnonymousRequestMatcher">
+            <argument />
```
dbu: this leads to the travis failure. i would define the service without argument at all.

dbu: hm, or maybe rather use a variable for the php session id - i think tihs is already asked for in the configuration, so we should use it here too.

bdunogier: > i think tihs is already asked for in the configuration

did you have something specific in mind ? I'm not sure what variable should actually be used there.

bdunogier:
Hmm, does anyone know what could cause `PHP Warning:  Unexpected character in input:  '' (ASCII=18) state=0 in /home/travis/build/FriendsOfSymfony/FOSHttpCacheBundle/vendor/symfony/symfony/src/Symfony/Bridge/Twig/Extension/CodeExtension.php on line 140` ?

Tests pass except for this error.

dbu:
the lowest version test fails. i just re-ran that test for the master branch and it works. can you please try rebasing this branch on master? or check the failures, if they are actually caused by a change in this branch?

dbu:
oh, just noticed this branch is against the 1.3 version. we now have 1.4 and should target 1.4 instead. or indeed master as its a new feature and there will be no 1.5 but 2 hopefully soonish.

bdunogier:
Has 1.3 reached end of life in between ? :)

dbu:
sorry, had lib and bundle open at the same time. the bundle is still at 1.3. so rebase on 1.3 should be the right thing.

bdunogier:
Note that iirc, I've done that yesterday.

```
$ git l origin/1.3..
79bbdb454b96da7ae46e557baa8834863b2d2034 fixup! Do not generate hash for anonymous requests
e4dca8a55c90ba62af278e5c149ffa59258e3b8a fixup! Do not generate hash for anonymous requests
a2a675baafbd4e189a1df293e25da8ac4b848978 fixup! Do not generate hash for anonymous requests
53764a2a146b68c4fe577862b676f464c1d39014 Do not generate hash for anonymous requests

$ git l ..origin/1.3
```

bdunogier:
@dbu is there anything I should do in particular to run the 1.3 tests ?

On both by branch and 1.3, after `composer update --prefer-lowest`:

```
$ SYMFONY_DEPRECATIONS_HELPER=weak ./vendor/bin/phpunit
... 
26) FOS\HttpCacheBundle\Tests\Functional\Test\ProxyTestCaseTest::testGetHttpClient
Class __PHP_Incomplete_Class has no unserializer

/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/vendor/symfony/symfony/src/Symfony/Component/Config/ConfigCache.php:76
/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:559
/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:132
/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Test/WebTestCase.php:43
/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/Test/ProxyTestCase.php:139
/Users/bdunogier/ezplatform/vendor/friendsofsymfony/http-cache-bundle/Tests/Functional/Test/ProxyTestCaseTest.php:20
```

dbu:
maybe you need to remove the cache files our little test symfony app generates?

bdunogier:
> maybe you need to remove the cache files our little test symfony app generates?

HAH, cache, I knew it !

Resources/config/user_context.xml
stof: it should be ``<argument type="collection" />`` to pass an empty array rather than an empty string

bdunogier:
One error left: 

```
PHP Warning:  Unexpected character in input:  '' (ASCII=18) state=0 in /home/travis/build/FriendsOfSymfony/FOSHttpCacheBundle/vendor/symfony/symfony/src/Symfony/Bridge/Twig/Extension/CodeExtension.php on line 140
```

It seems to be [fixed in Sf 2.3.4](https://github.com/symfony/symfony/commit/e8e76ece5aeed8c3ca7da0741da9d93bc0ec5757), but the failing test uses 2.3.0 (`--prefer-lowest`). My main doubt is that it apparently doesn't happen without this PR, but I hardly see how it can cause this.

Hints, anybody ? @stof maybe ?

lolautruche:
+1 with the solution

Shouldn't there be some semantic config in order to be able to configure session prefix?

bdunogier:
> Shouldn't there be some semantic config in order to be able to configure session prefix?

I see that as a low level feature, where you override the anonymous request checker when you need to. Given that it hasn't been reported much, I don't see it as a very common use-case.

By the way, one *question*: I was thinking about renaming the `AnonymousRequestMatcher` class to `SessionNamePrefixRequestMatcher`, so that the name actually says what it is. The service's name would remain the same, for the same reason.

+1 ? -1 ?

bdunogier:
Actually, I've managed to get that error by running phpunit using the `.phar`. I get an s-load of those unexpected character in input. It doesn't explain why it passes on travis, though...

bdunogier:
What remains is this:

```
1) FOS\HttpCacheBundle\Tests\Unit\EventListener\UserContextSubscriberTest::testFullAnonymousRequestHashNotGenerated
BadMethodCallException: Method Mockery_16_FOS_HttpCache_UserContext_HashGenerator::shouldNotReceive() does not exist on this mock object
```

I'm a bit skeptical, since the mock is created exactly like in the method above, and the method should exist. But I'll dig a bit to fix it:

```
        $hashGenerator = \Mockery::mock('\FOS\HttpCache\UserContext\HashGenerator');
        $hashGenerator->shouldNotReceive('generateHash');
```

dbu:
if sf 2.3.4 fixes the weird encoding problem, lets just raise our version constraint to `"^2.3.4||^3.0"`. the point of providing symfony 2.3 support is that people can still get updates. but if they don't update symfony itself to the latest 2.3, they don't deserve to get other updates ;-)

dbu:
imo things look sane. is there any chance of a security issue if we don't sanity check the hash when the user is anonymous? the worst case i can think of is when the caching proxy still sees a session cookie and uses an authenticated hash, but symfony does not see the session and our check would treat this request as anonymous and not detect the mismatch. then it could happen that the anonymous page is cached for authenticated users of the users previous groups. which is at least no security problem, just annoying. but i am not sure if it can ever happen - if the session cookie is still in the request, we would not see the request as anonymous, even if it actually is, right?

bdunogier:
> if sf 2.3.4 fixes the weird encoding problem, lets just raise our version constraint to "^2.3.4||^3.0". the point of providing symfony 2.3 support is that people can still get updates. but if they don't update symfony itself to the latest 2.3, they don't deserve to get other updates ;-)

Just tried: it does. I only have my `shouldNotReceive()` error, but the encoding errors are gone. I'll add a commit with that change to the PR.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,69 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Matches anonymous requests.
+ *
+ * Uses HTTP authentication headers as well as session name matching.
+ */
+class AnonymousRequestMatcher implements RequestMatcherInterface
```
bdunogier: Yes, I will.

bdunogier:
Unit test for `AnonymousRequestMatcher` added.

bdunogier:
Removed the `shouldNotReceive()` call from the method added to `UserContextSubscriberTest`.

According to the debugger, it has the `shouldReceive` method, but _not_ `shouldNotReceive`. I'm a bit skeptical, but...

bdunogier:
Poke @dbu. As far as I know, the review remarks have been addressed. Let me know if something is missing. If not, I'll squash the commits so that you can merge this.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,69 @@
...
+        $resolver->setDefaults(['session_name_prefix' => 'PHPSESSID']);
+
+        $this->options = $resolver->resolve($options);
+    }
+
+    public function matches(Request $request)
+    {
+        if ($request->server->has('AUTHORIZATION') ||
+            $request->server->has('HTTP_AUTHORIZATION') ||
+            $request->server->has('PHP_AUTH_USER')
```
dbu: what if people use their custom authentication scheme?

dbu: i noticed that we have a configuration `user_identifier_headers`. maybe we should instead of looking for a session cookie or these headers just inject that list of headers and check for those headers.

the idea with cookies is that as we Vary anyways, the caching proxy has to clean up the cookie to not have random js cookie things in it. so we would not need to really go into the individual cookies.

dbu:
sorry that i missed this. had a look again, and i think we are almost there. looking over the code one more time, i noticed that we have a inconsistency with what we use for the Vary header. can you look at that please?

i was thinking whether we need any documentation, but its probably too low level.

UserContext/AnonymousRequestMatcher.php
dbu: does symfony itself look at HTTP_AUTHORIZATION and PHP_AUTH_USER? if so, we should change the default value of user_identifier_header in the config.

bdunogier:
> i was thinking whether we need any documentation, but its probably too low level.

Honestly glad you say that :-)
But if you think it is worth it, I can cut-my-own-throat and write something down.

> i noticed that we have a inconsistency with what we use for the Vary header. can you look at that please?

Sure, I'll have a look, thank you.

UserContext/AnonymousRequestMatcher.php
bdunogier: > the idea with cookies is that as we Vary anyways, the caching proxy has to clean up the cookie to not have random js cookie things in it. so we would not need to really go into the individual cookies.

Meaning that a request is anonymous only if it doesn't _have_ any of the headers from `user_identifier_headers`. The part that checks for custom sessions names should then be removed completely, right ?

bdunogier: Note that `HTTP_AUTHORIZATION` and `PHP_AUTH_USER` are server variables, not cookies. The cookie would, afaik, be Authorization in both cases.

bdunogier:
> i noticed that we have a inconsistency with what we use for the Vary header. can you look at that please?

@dbu that's about the `user_identifier_headers` right ? Or did you have something else in mind ?

UserContext/AnonymousRequestMatcher.php
dbu: > The part that checks for custom sessions names should then be removed completely, right ?

yes, exactly. because we also Vary on those headers, if you send a cookie header with google analytics data in it, you break the cache anyways. this has to be handled on the caching proxy.

it looks like header based authentication is always put into the AUTHORIZATION header: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/ServerBag.php#L94-L98 so cookie and authorization should be enough

dbu:
yes, that was about the comment on the code we discuss. just wanted to summarize what remains open to do.

UserContext/AnonymousRequestMatcher.php
andrerom: @dbu so to double check here: in default vcl we strip every other cookie other then session. And from what you say this is done somewhere with the Sf proxy as well, so essentially here if there is *any* cookie, it is session cookie meaning we don't need to loop true them here. Correct?

bdunogier:
Changed the AnonymousRequestMatcher:
- receives the `user_identifier_headers`
- checks for the given headers (the part specific to cookies should maybe be changed, please let me know)

UserContext/AnonymousRequestMatcher.php
dbu: that is what i think. i just double-checked and for HttpCache, this is in https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/UserContextSubscriber.php#L124-L137

for varnish, this is documented here: http://foshttpcache.readthedocs.io/en/latest/varnish-configuration.html#cleaning-the-cookie-header

caching the user context hash without cleaning the cookie header can not work.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,42 @@
...
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
+/**
+ * Matches anonymous requests using a list of identification headers.
+ */
+class AnonymousRequestMatcher implements RequestMatcherInterface
+{
+    private $userIdentifierHeaders;
+
+    public function __construct($userIdentifierHeaders = ['Cookie', 'Authorization'])
```
dbu: as this is in the symfony bundle and always created from the service container, i don't think we need to default this. the Configuration.php already defines defaults for user_identifier_headers.

and please typehint array for $userIdentifierHeaders and add a docblock that says that its a list of the header names of headers that make a request not be anonymous.

dbu:
apart from https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/289#discussion_r76057700 this looks ready to merge. some build fail because composer has too many options and runs out of memory. not related to this PR

UserContext/AnonymousRequestMatcher.php
bdunogier: Ack. Note that `UserContextSubscribed` does set a default ctor value for them.

bdunogier:
Changed. Let me know if it's okay, and I'll cleanup the commits a bit.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,45 @@
...
+ * Matches anonymous requests using a list of identification headers.
+ */
+class AnonymousRequestMatcher implements RequestMatcherInterface
+{
+    private $userIdentifierHeaders;
+
+    /**
+     * @param array $userIdentifierHeaders List of request headers that authenticate a non-anonymous request
+     */
+    public function __construct(array $userIdentifierHeaders = [])
```
dbu: my idea was to not specify a default at all, so that calling code is forced to pass an argument and not null. as it is now its is rather dangerous, if anything would go wrong we would simply consider everything anonymous.

dbu:
apart from that 1 comment i am happy with this. please squash the commits as you see fit, and rebase on 1.3 to get the fixed testing setup i just merged.

UserContext/AnonymousRequestMatcher.php
bdunogier: Removed the default altogether, and updated the unit test.
Rebased on top of 1.3, and cleaned the commits up.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->userIdentifierHeaders = $userIdentifierHeaders;
+    }
+
+    public function matches(Request $request)
+    {
+        foreach ($this->userIdentifierHeaders as $header) {
+            if (strtolower($header) === 'cookie' && $request->cookies->count()) {
+                return false;
+            }
+            if ($request->headers->has($header)) {
```
dbu: sorry, one more question: what does symfony request do with the cookie header? is that removed from the header bag? what happens if we have a request with only an empty cookie header? won't we then get in here and see the cookie header, making the previous if on the cookies count kind of pointless? should we do

```
if ($request->headers->has($header)) {
    if (strtolower($header) === 'cookie' && 0 === $request->cookies->count()) {
        // ignore empty cookie header
        continue;
    }

    return false;
}
```

bdunogier: Done, with test, @dbu.

bdunogier:
By the way, once this is merged, do you think we'll be able to tag a new release shortly, @dbu ?

UserContext/AnonymousRequestMatcher.php
dbu: the first if with cookie header and nonzero number of cookies is now redundant and should be removed.

bdunogier: Removed the if, adjusted the existing test, and added a new case.

dbu:
yay, thanks bertrand

dbu:
ported to master as well, and tagged 1.3.8

blankse:
I have a problem with this.

I use eZ Platform and a anonymous with cookies (e. g. from piwik) but without the session cookie doesn't use the http cache.

Why would only look after cookie count and not after the specific session cookie?

blankse:
I created a PR for my issue: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/421

PR #295, opened 2016-07-01
Updated Symfony HTTP cache docs

dbu:
Warning about class caches in Symfony 2.8+.

### DIFF
--- a/Resources/doc/features/symfony-http-cache.rst
+++ b/Resources/doc/features/symfony-http-cache.rst
@@ -76,6 +76,26 @@ To register subscribers that you need to instantiate yourself, overwrite
         return $subscribers;
     }
 
+.. warning::
+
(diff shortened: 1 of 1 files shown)


dbu:
thanks a lot @pestaa !

glye:
Hi @dbu & @pestaa! I'm trying the workaround of untagging the class cache warmer in a compiler pass. It is working per se, but still fails because Symfony's AddCacheWarmerPass runs before mine, so my untagging happens too late. Is it possible to preempt this? Many thanks for any help.

glye:
Nevermind, I found a way: Instead of untagging the class cache warmer, I read the warmers already loaded by Symfony, filter out the warmer I don't want, and rewrite the list of warmers, like AddCacheWarmerPass does:
`$container->getDefinition('cache_warmer')->replaceArgument(0, $warmers);`
Feels a bit hackish, but it works.

PR #304, opened 2016-08-04
chore: add phpunit to require-dev for testing

gsdevme:
I made a similar to https://github.com/FriendsOfSymfony/FOSHttpCache/pull/290

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,7 @@ before_script:
   - composer update $COMPOSER_FLAGS
 
 script:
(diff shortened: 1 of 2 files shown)


dbu:
hm. composer now explodes with out of memory. likely this has nothing to do with this PR but with symfony 3.1 being released and thus we have even more permutations of options to install. restarted a master build to see what happens there

dbu:
yep, seems to be we have too many options. will do a PR to fix that. thanks for this cleanup here

PR #310, opened 2016-08-25
test with symfony 3.1

dbu:


dbu:
blocked by https://bugs.php.net/bug.php?id=72906 - lets wait for beta4/stable release where this is supposed to be fixed.

PR #311, opened 2016-09-05
handle configuration where we only have a custom proxy client and proxies configured

dbu:
without this fix, a config like the following leads to an error during container building:

```
    cache_manager:
        custom_proxy_client: "app.caching.rabbitmq_client"
```

`Undefined index: proxy_client`

### DIFF
--- a/DependencyInjection/FOSHttpCacheExtension.php
+++ b/DependencyInjection/FOSHttpCacheExtension.php
@@ -64,7 +64,7 @@ class FOSHttpCacheExtension extends Extension
         }
 
         if ($config['cache_manager']['enabled']) {
-            if (!empty($config['cache_manager']['custom_proxy_client'])) {
+            if (array_key_exists('custom_proxy_client', $config['cache_manager'])) {
                 // overwrite the previously set alias, if a proxy client was also configured
                 $container->setAlias(
                     $this->getAlias().'.default_proxy_client',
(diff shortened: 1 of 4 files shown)


Resources/doc/reference/configuration/cache-manager.rst
```diff
@@ -39,6 +39,10 @@ your own service that implements ``FOS\HttpCache\ProxyClientInterface``.
         cache_manager:
             custom_proxy_client: acme.caching.proxy_client
 
+When you specify a custom proxy client, the bundle does not know about the
+capabilities of the client. The ``generate_url_type`` defaults to true and
+:doc:`tag support <tags>` is only active if explicitly enabled.
+
 ``generate_url_type``
```
Tobion: The constants of `UrlGeneratorInterface` changed in symfony 3 to integer instead of string/bool. So the documentation about those enum values is not correct anymore.

DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -313,14 +319,20 @@ private function loadSymfony(ContainerBuilder $container, XmlFileLoader $loader,
         }
     }
 
+    /**
+     * @param ContainerBuilder $container
+     * @param XmlFileLoader    $loader
+     * @param array            $config    Configuration section for the tags node
+     * @param string|bool      $client    Name of the client or false when a custom client is used
```
Tobion: `string|false`. But I don't like multi-type params. So I'd use the empty string instead of false.

Tobion:
:+1:

PR #316, opened 2016-09-29
Fixed documentation

nico-incubiq:
Fixed the documentation concerning the option value for disabling the UserContextSubscriber


dbu:
thanks. i think using the XOR operator ^ could work here as well, but your fix is correct.

PR #319, opened 2016-11-24
cleanup event listener naming

### LINKED ISSUE #190: fix naming of event listeners
we provide a couple of event listeners: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/tree/master/EventListener

according to naming convention, those [should end with the word listener](http://symfony.com/doc/2.3/cookbook/bundles/best_practices.html#classes), not subscriber. they do implement the event subscriber interface, but their purpose is to be event listeners.

i suggest we do that when we get ready for a version 2 - its not that important and its a BC break for people extending the classes or doing things with the services.

dbu:
fix #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Changelog
 * [User Context] Added an option always_vary_on_context_hash to make it 
   possible to disable automatically setting the vary headers for the user 
   hash.
+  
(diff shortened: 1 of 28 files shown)


dbu:
@ddeboer would love to merge this before attacking #235 again, to keep things apart. this is quite a lot of editing on its own.

ddeboer:
Cheers!

PR #320, opened 2016-11-24
use class name constants instead of strings for mocks

dbu:


Tests/Unit/EventListener/InvalidationListenerTest.php
```diff
@@ -39,12 +48,6 @@ public function testNoRoutesInvalidatedWhenResponseIsUnsuccessful()
             ->shouldReceive('invalidateRoute')->never()
             ->shouldReceive('flush')->once();
 
-        $this->invalidators = \Mockery::mock('\FOS\HttpCacheBundle\Invalidator\InvalidatorCollection')
-            ->shouldReceive('hasInvalidatorRoute')
-            ->with('my_route')
-            ->andReturn(false)
-            ->getMock();
```
dbu: this was completely unused and there is no InvalidatorCollection class. i assume this was a leftover from development that we solved differently in the end.

PR #322, opened 2016-11-29
we need stdlib.h for the reverse-proxy-ttl handling

dbu:
@ddeboer just noticed this in one of our projects. and we have no functional test for this one... if you have ideas how to improve the text below the code example, they would be welcome.

### DIFF
--- a/Resources/doc/reference/configuration/headers.rst
+++ b/Resources/doc/reference/configuration/headers.rst
@@ -336,6 +336,10 @@ you need to extend your varnish ``vcl_fetch`` configuration:
 
 .. code-block:: c
 
+    C{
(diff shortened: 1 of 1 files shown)


ddeboer:
Makes sense. I guess for the 2.0 we can just refer to [the library](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/d7b79e00abe32de6829bc43033f44e0303a05df1/resources/config/varnish/fos_custom_ttl.vcl)?

dbu:
ah indeed. #235 will remove this again. from the library, it seems that varnish 4 needs this, whereas varnish 3 did not need it (i guess it was just available automatically)

Issue #323, opened 2016-12-01
Deprecation warning with Symfony 3.2


COil:
User Deprecated: Checking only for cacheable HTTP methods with Symfony\Component\HttpFoundation\Request::isMethodSafe() is deprecated since version 3.2 and will throw an exception in 4.0. Disable checking only for cacheable methods by calling the method with `false` as first argument or use the Request::isMethodCacheable() instead.

	  /EventListener/CacheControlSubscriber.php:223: {▼
	    {
	         return $request->isMethodSafe();
	     }
	  }

PR #324, opened 2016-12-01
Deprecation

COil:
See #323

### DIFF
--- a/EventListener/CacheControlListener.php
+++ b/EventListener/CacheControlListener.php
@@ -220,6 +220,6 @@ class CacheControlListener extends AbstractRuleListener implements EventSubscrib
      */
     protected function isRequestSafe(Request $request)
     {
-        return $request->isMethodSafe();
+        return $request->isMethodCacheable();
     }
 }


stof:
shouldn't this require dropping support for some Symfony versions though, regarding the release containing this new method ?

COil:
It was introduced in 3.2 indeed...

ddeboer:
Thanks for this PR, @COil!

> shouldn't this require dropping support for some Symfony versions though, regarding the release containing this new method ?

@stof Wouldn’t it be better to do feature discovery, so to check if the `isMethodCacheable()` method exists (>= 3.2) and otherwise call `isMethodSafe()` (< 3.2)?

stof:
@ddeboer as the method was also added in older branches in a patch release (to fix spec compliance), I would just drop support for early patch versions. this would avoid the need to maintain a BC layer

The constraint on HTTPFoundation would be ``~2.7.0 || ~2.8.13 || ^3.1.6``.

ddeboer:
> @ddeboer as the method was also added in older branches in a patch release

Ah, didn’t know that. In that case, @COil please change the version constraint.

dbu:
i like what stof suggested in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/324#issuecomment-264452429 - @COil can you please adjust the dependencies?

it looks like symfony 3.0 did not get the change retrofitted - is that correct @stof or has it just not been tagged yet?

note that master is the upcoming 2.0 and there is a bunch of things to wrap up before we can release it. we could also do this change in the 1.3 branch but there we would need checks whether the method exists to also support symfony 2.3. if you want the fix soon @COil it would be great if you can do that first, as PR against 1.3 and then i merge 1.3 to master and you can apply this change.

jderusse:
You should also fix the [TagListener](https://github.com/COil/FOSHttpCacheBundle/blob/e0780ba214cb2054aa7c24ad10dbcbda1dc71d6c/EventListener/TagListener.php#L82).

dbu:
ping @COil

xabbuh:
@dbu Symfony 3.0 reached EOM before this patch was applied. So you will then have to drop support for it.

dbu:
i guess dropping sf 3.0 support in the 2.0 bundle is all right then. people that move to fos httpcache 2 will also keep their symfony up to date, i hope.

ddeboer:
So, @COil, please

- change the version constraint as described above (excluding Symfony 3.0)
- rebase this on master because there seems to be a merge conflict – though I’m not sure why :wink:

Thanks!

COil:
hello, @stof can you do it please ? 👍

Tobion:
> The constraint on HTTPFoundation would be `~2.7.0 || ~2.8.13 || ^3.1.6`.

@stof the contraints actually needs to be `~2.7.20 || ~2.8.13 || ^3.1.6`

j0k3r:
@COil can't you do the rebase on your own? :thinking:

COil:
Don't have much time actually, can you do it please ?

Tobion:
See #337 for 1.3.

ddeboer:
Replaced with #336.

PR #325, opened 2016-12-03
Use HashGeneratorInterface

Nyholm:
See https://github.com/FriendsOfSymfony/FOSHttpCache/pull/336

### DIFF
--- a/EventListener/UserContextListener.php
+++ b/EventListener/UserContextListener.php
@@ -11,7 +11,7 @@
 
 namespace FOS\HttpCacheBundle\EventListener;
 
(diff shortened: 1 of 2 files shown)


dbu:
you also need to adjust what we mock, it seems. but with https://github.com/FriendsOfSymfony/FOSHttpCache/pull/335 the naming will change.

Nyholm:
hm, Im closing this PR. Making `HashGenerator` an interface in the cache library does not effect this bundle.

dbu:
yay for unit tests :-)

PR #329, opened 2016-12-06
improve file structure

dbu:


ddeboer:
@dbu Just noticed this change because the Readthedocs build is failing (which is easily fixed). Could you please explain the rationale for this change? While I like the cleaner structure, are there any ‘official’ recommendations to do it like this? E.g. the [Best Practices](https://symfony.com/doc/current/bundles/best_practices.html) still have the `Resources/doc` dir.

dbu:
i was not aware of the best practices for this one, sorry. fixing in #339

PR #333, opened 2016-12-07
Fix service definition for new Symfony parent class (HttpProxyClient)

ahilles107:


dbu:
ups, this is scary. we seem to lack a test that highlights this problem, the build should not be green with this. thanks for the fix! created #334 as a reminder that we need to add some tests.

PR #336, opened 2016-12-22
Symfony 3.2

j0k3r:
Should replace https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/324

Fix deprecated methods
- Add support for Symfony 3.2
- Fix the default private cache control added by Symfony 3.2
- Use `isMethodCacheable` instead of `isMethodSafe` (because it's deprecated)

I hope the new `symfony/http-foundation` restriction is ok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,6 @@ matrix:
         - SYMFONY_VERSION='2.8.*'
     - php: 7.1
       env:
-        - SYMFONY_VERSION='3.0.*'
-    - php: 7.1
-      env:
         - SYMFONY_VERSION='3.1.*'
     - php: 7.1
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ matrix:
     - php: 7.1
       env:
         - SYMFONY_VERSION='3.1.*'
+    - php: 7.1
+      env:
+        - SYMFONY_VERSION='3.2.*'
 install:
   - pip install -qr doc/requirements.txt --user
 
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
     "require": {
         "php": "^5.5.0 || ^7.0.0",
         "friendsofsymfony/http-cache": "2.0.x-dev",
-        "symfony/framework-bundle": "^2.8||^3.0"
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -23,7 +23,8 @@
     "require": {
         "php": "^5.5.0 || ^7.0.0",
         "friendsofsymfony/http-cache": "2.0.x-dev",
-        "symfony/framework-bundle": "^2.8||^3.0"
+        "symfony/framework-bundle": "^2.8||^3.0",
+        "symfony/http-foundation": "~2.8.13 || ^3.1.6 || ^3.2.0"
```
Tobion: the `^3.2.0` is useless as that is included in the `^3.1.6`

.travis.yml
```diff
@@ -47,6 +47,9 @@ matrix:
     - php: 7.1
       env:
         - SYMFONY_VERSION='3.1.*'
+    - php: 7.1
+      env:
+        - SYMFONY_VERSION='3.2.*'
```
Tobion: the test against symfony 3.0 should be removed now

composer.json
```diff
@@ -23,7 +23,8 @@
     "require": {
         "php": "^5.5.0 || ^7.0.0",
         "friendsofsymfony/http-cache": "2.0.x-dev",
-        "symfony/framework-bundle": "^2.8||^3.0"
+        "symfony/framework-bundle": "^2.8||^3.0",
+        "symfony/http-foundation": "~2.8.13 || ^3.1.6"
```
ddeboer: 👍

ddeboer:
Thanks for fixing this!  😄

composer.json
stof: even 2.7 has a version with the new method IIRC

stof: well, the bundle requires FrameworkBundle 2.8 though, so it is fine

iJanki:
Can we have a tagged version with this fix?

j0k3r:
@iJanki what you are looking for is https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/338 (and I have one last thing to fix before it can be merged)

dbu:
i just tagged 1.3.10 with #338 in it

PR #337, opened 2016-12-22
Use new Request::isMethodCacheable when available 

Tobion:
Request::isMethodSafe has different semantics and is deprecated for the old use-case.

Replaces #324 for 1.3


j0k3r:
There were at least 2 others places where `isMethodSafe` is used. Can I submit a new PR including your commit and my changes?

Tobion:
Sure.

PR #338, opened 2016-12-22
Add support for Symfony 3.*

j0k3r:
Following https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/337 & https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/324

I've added Symfony 3.1 & 3.2 to Travis.
~~There seems to be a strange error on Symfony 3.1 with a _missing_ Twig template ..~~
It seems to appear only on my dev machine.

### DIFF
--- a/EventListener/AbstractRuleSubscriber.php
+++ b/EventListener/AbstractRuleSubscriber.php
@@ -54,4 +54,16 @@ class AbstractRuleSubscriber
 
         return false;
     }
+
+    /**
+     * Decide whether to even look for matching rules with the current request.
+     *
+     * @param Request $request
+     *
+     * @return bool True if the request is safe and headers can be set
+     */
+    protected function isRequestSafe(Request $request)
+    {
+        return method_exists($request, 'isMethodCacheable') ? $request->isMethodCacheable() : $request->isMethodSafe();
+    }
 }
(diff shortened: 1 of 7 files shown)


Tests/Functional/Fixtures/Controller/TagController.php
```diff
@@ -19,6 +19,18 @@
 class TagController extends Controller
 {
     /**
+     * Workaround to be compliant with all Symfony version.
+     *
+     * @param Request $request
+     *
+     * @return bool True if the request is safe
+     */
+    protected function isRequestSafe(Request $request)
```
Tobion: should be private. otherwise its a new feature

Tobion:
👍

EventListener/AbstractRuleSubscriber.php
```diff
@@ -54,4 +54,16 @@ protected function matchRule(Request $request, Response $response)
...
     }
+
+    /**
+     * Decide whether to even look for matching rules with the current request.
+     *
+     * @param Request $request
+     *
+     * @return bool True if the request is safe and headers can be set
+     */
+    protected function isRequestSafe(Request $request)
```
dbu: can we rename this method to `isRequestCacheable`? this is what we semantically want to know.

j0k3r:
@dbu I've renamed the method.
About the matrix & php 7, maybe we should replace php 5.6 in the matrix by php 7 (otherwise we can duplicate all the matrix for php 7 but the build might take ages to end after ...)

dbu:
thank you. i think we should not bother changing the matrix for the 1.3 branch anymore. the master which will be 2.0 is using php 7.1 atm.

PR #340, opened 2017-01-06
fix license year and version in doc

dbu:


Resources/doc/conf.py
```diff
@@ -49,16 +49,16 @@
...
-copyright = '2014-2015, David de Boer, David Buchmann'
+copyright = '2014-2017, David de Boer, David Buchmann'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = '1.3.0'
+version = '2.0.0'
```
dbu: i guess this makes more sense, even if 2 is not released yet

PR #341, opened 2017-01-06
merge 1.3 to master

dbu:
removing a lot that was added in 1.3 to support both older and newest versions of symfony correctly.

apart from getting some fixes in, this will make sure we can merge 1.3 to master again if there are further fixes to 1.3.

### DIFF
--- a/src/EventListener/CacheControlListener.php
+++ b/src/EventListener/CacheControlListener.php
@@ -102,7 +102,7 @@ class CacheControlListener extends AbstractRuleListener implements EventSubscrib
         }
 
         // do not change cache directives on unsafe requests.
-        if ($this->skip || !$this->isRequestSafe($request)) {
+        if ($this->skip || !$request->isMethodCacheable()) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/EventListener/CacheControlListener.php
```diff
@@ -210,16 +210,4 @@ private function setExtraCacheDirectives(Response $response, array $controls, $o
...
     }
-
-    /**
-     * Decide whether to even look for matching rules with the current request.
-     *
-     * @param Request $request
-     *
-     * @return bool True if the request is safe and headers can be set
-     */
-    protected function isRequestSafe(Request $request)
```
dbu: imo we don't need this method anymore when there is no BC method_exists stuff anymore

stof: why was it protected in 1.3 btw ?

dbu: that looks like a mistake, it should have been private. the listener is not supposed to be extended.

PR #349, opened 2017-02-05
Add tags.strict config param

### LINKED ISSUE #278: configure strict mode
as a result of https://github.com/FriendsOfSymfony/FOSHttpCache/issues/270 being solved & merged, have the ability to configure strict mode.

will have a go at this somewhere this week.
question beforehand: any ideas on where to put this in the configuration? currently strict mode only applies to tagging, but i can imagine it can be applied to other (future) validations in the library…

ddeboer:
Fix #278


dbu:
thanks!

PR #351, opened 2017-02-05
Use OptionsResolver in event listener

### LINKED ISSUE #290: use OptionsResolver where it makes sense
- event subscribers
- check for other classes

ddeboer:
Fix #290.


ddeboer:
This was the only place that seems to make sense for OptionsResolver. Or do we also want to refactor The UserContextListener constructor to accept an array of options instead of separate arguments?

dbu:
looking at the number of optional parameters in that event listener, i think we should refactor it. looking at the code, i got confused by https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/UserContextListener.php which in a way looks very similar while being something very different. i wonder if we can/should do something about that. created #352

either way, this can be done as separate pull request, so i will merge this.

Issue #286, opened 2016-02-24
Cache tags not being added on redirect response.


thisispaul:
The `X-Cache-Tags` header is not being added to Redirect Responses and I think it should be. I have a service which generates screenshots based on some data and I need to cache the redirect to the file it generates and then invalidate when the data is updated. Tags seem to be the way to do this however the tags are not added on redirect responses.

The offending code is on line 70 [TagSubscriber.php](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/TagSubscriber.php#L70) where it checks to see if the response is successful

```
        if ($response->isSuccessful()) {
            $tags = $this->getAnnotationTags($request);
        }
```

I'm proposing a simple change to this so it also tags redirect responses.

```
        if ($response->isSuccessful() || $response->isRedirection()) {
            $tags = $this->getAnnotationTags($request);
        }
```

Is there any reason this isn't like this already/a different way to get it to tag redirect response (HTTP Code 302).

I'm happy to submit a pull request if this is fine.

dbu:
this request is somewhat similar to #279. @wickedOne said he wanted to work on #69 which would solve this by using matchers to decide what tags and what triggers invalidation.

for tagging the response, i am actually not sure why we don't tag everything. usually, error pages or temporary redirects will not be cached at all. but if they are, a change that triggers a tag invalidation might also fix the error, so there is no harm in tagging them.

@ddeboer what is your opinion on this? why should we not just tag every response regardles of status? with the current code, a redirect response to a POST would also not invalidate the tags... we did have the discussion to not invalidate on fail responses - though if we make things more flexible here, we might want the default more agressively invalidating. it is better to err on the side of invalidating too much than too little.

ddeboer:
I’m in favour of removing the `$response->isSuccessful()` check when *setting* tags, as it will make the behaviour more predictable. When *invalidating* tags, we can do so for redirect responses.

However, I’m not so sure about invalidating in case of error responses. When we start invalidating tags even in case of 400 errors (for instance), it becomes to easy for clients to swamp the server. 

What about using the same logic as we have in [InvalidationSubscriber](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/InvalidationSubscriber.php#L92-L93)?

ddeboer:
Fixed in #354.

Issue #279, opened 2016-02-02
[cache tagging] OPTIONS call invalidates tags


wickedOne:
currently, when doing an [`OPTIONS`](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.2) call to a specific route which is configured for tagging, [this code](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/TagSubscriber.php#L82-L91) invalidates the related / generated tags.

i'm aware that you can easily circumvent this behaviour by adding a match like this to the top of your config:

``` yml
-
  match:
      path: ^/
      methods:
          - OPTIONS
```

but am wondering, looking at the specifications for the `OPTIONS` call, whether this is desired behaviour

dbu:
hm, interesting. i guess options could usually even be cached. but probably the HTTP specification does not guarantee this, so its considered not safe.

if memory serves correctly, #69 would provide a way to solve this (see also #241) by offering more flexibility.

ddeboer:
Fixed in #354.

Issue #241, opened 2015-10-09
Taging via configuration doesn't work for BAN requests


tomaszhanc:
I add tags via YAML file, with configuration similiar to this one:

``` yml
    fos_http_cache:
        tags:
            rules:
                -
                    match:
                        path: ^/config
                    tag: [config]
```

It works great for responses, but it doesn't add these tags for BAN request to Varnish after POST method which is returning a status of 201. I checked adding tags via annotation and it works ok there. I've found out that these lines in RuleMatcher are causing the problem:

``` php
    $status = array(200, 203, 204, 300, 301, 302, 404, 410);
    if (!empty($this->criteria['additional_cacheable_status'])) {
        $status = array_merge($status, $this->criteria['additional_cacheable_status']);
    }
    if (!in_array($response->getStatusCode(), $status)) {
        return false;
    }
```

Adding to confgiuration option `additional_cacheable_status: [201]` solves the problem, but it's a bit tedious to do for every rule. Can we add this status code as a default?

dbu:
hm, looks like cacheable status is not exactly the same as invalidatable status... we should not cache responses with status 201 but invalidating sounds right.

if you see an idea how to distinguish the two things without making everything very complicated, please do a pull request against master.

teohhanhui:
#230 only fixed it for `invalidation` rules (config). See #69. It seems to be a longstanding issue...

dbu:
yeah nobody got around to implement #69 so far, unfortunately. if anybody has time to contribute that, would be awesome!

ddeboer:
Fixed in #354.

PR #354, opened 2017-02-07
Add rule matchers for cacheable and must invalidate

### LINKED ISSUE #69: RuleMatcher for TagSubscriber, InvalidationSubscriber

ddeboer:
Fix #69, #241, #279, #286.

This is a proposal for fixing #69. It’s a WIP but please have a look before I continue: do we agree on the direction taken here? 

The idea is to:

1. move the logic for determining ‘cacheable’ and [‘must invalidate’](https://tools.ietf.org/html/rfc7234#section-4.4) into rule matchers and then re-use those between the listeners and any other classes that need the logic
2. expose those rule matchers as services for customising (e.g.) cacheable status codes
3. get rid of the `additional_cacheable_status` per rule: do we really need that or can we suffice with one globally defined set of cacheable status codes?


tests/Functional/Fixtures/Controller/TagController.php
```diff
@@ -69,6 +69,6 @@ public function subrequestAction()
 
     public function twigAction()
     {
-        $this->render('::tag.html.twig');
+        return $this->render('::tag.html.twig');
```
ddeboer: Not sure how this worked before: a controller must return a response.

src/EventListener/TagListener.php
```diff
@@ -74,12 +89,14 @@ public function onKernelResponse(FilterResponseEvent $event)
         $request = $event->getRequest();
         $response = $event->getResponse();
 
-        $tags = [];
-        // Only set cache tags or invalidate them if response is successful
-        if ($response->isSuccessful()) {
-            $tags = $this->getAnnotationTags($request);
+        if (!$this->cacheableRule->matches($request, $response)
+            && !$this->mustInvalidateRule->matches($request, $response)
```
dbu: this would be an odd situation when neither matches. but i think there is not much else we can do.

src/EventListener/TagListener.php
```diff
@@ -88,14 +105,15 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+        if ($this->cacheableRule->matches($request, $response)) {
+            // For safe requests (GET and HEAD), set cache tags on response
             $this->symfonyResponseTagger->addTags($tags);
             if (HttpKernelInterface::MASTER_REQUEST === $event->getRequestType()) {
-                // For safe requests (GET and HEAD), set cache tags on response
                 $this->symfonyResponseTagger->tagSymfonyResponse($response);
             }
-        } elseif (count($tags)) {
-            // For non-safe methods, invalidate the tags
+        } elseif (count($tags)
```
dbu: i would prfer to return early if !count($tags). with your early return if its neither cacheable nor mustinvalidate, this here would be a simple `else` i think.

src/Http/ResponseMatcher/CacheableResponseMatcher.php
```diff
@@ -0,0 +1,42 @@
...
+ * @see https://tools.ietf.org/html/rfc7231#section-6.1
+ */
+class CacheableResponseMatcher implements ResponseMatcherInterface
+{
+    private $cacheableStatusCodes = [
+        200, 203, 204, 206,
+        300, 301,
+        404, 405, 410, 414,
+        501,
+    ];
```
dbu: isn't this list defined in symfony itself somewhere? do we need to redo it?

src/Http/ResponseMatcher/CacheableResponseMatcher.php
```diff
@@ -0,0 +1,42 @@
...
+
+    private $additionalStatusCodes;
+
+    public function __construct(array $additionalStatusCodes = [])
+    {
+        $this->additionalStatusCodes = $additionalStatusCodes;
```
dbu: we could merge the additional codes with the default codes to make the check below simpler.

src/Http/ResponseMatcher/ResponseMatcher.php
```diff
@@ -0,0 +1,63 @@
...
+use Symfony\Component\DependencyInjection\ExpressionLanguage;
+use Symfony\Component\HttpFoundation\Response;
+
+class ResponseMatcher implements ResponseMatcherInterface
+{
+    private $statusCodes;
+    private $expression;
+    private $expressionLanguage;
+
+    public function matchExpression($expression)
```
dbu: i would prefer to call this setMatchExpression to make it explicit this is setting up the matcher. (and same for status code below). or have a constructor so that we can require at least one of the two criteria to exist? 

the phpdoc for $expression should explain what context the expression gets.

tests/Functional/Fixtures/Controller/TagController.php
dbu: it probably means the test does not check the returned response as that will be a 500 status...

tests/Unit/EventListener/TagListenerTest.php
```diff
@@ -48,7 +56,22 @@ public function setUp()
...
 
-        $this->listener = new TagListener($this->cacheManager, $this->symfonyResponseTagger);
+        $this->cacheableRule = new RuleMatcher(
+            new CacheableRequestMatcher(),
+            new CacheableResponseMatcher()
+        );
+
+        $this->mustInvalidateRule = new RuleMatcher(
+            new UnsafeRequestMatcher(),
+            new NonErrorResponseMatcher()
```
dbu: strictly speaking, a unit test should mock the rule matcher for the behaviour we expect and then have separate unit tests for each matcher...

ddeboer:
> i think instead of having additional cacheable states globally or specifically, we should document which services to replace when you want to change the rules. 

That’s fine with me. What about ditching `ResponseMatcher` (that takes an expression or status code) and *only* having `NonErrorResponseMatcher` (invalidates) and `CacheableResponseMatcher` (set caching headers/tags)? Would people still need the `additional_cacheable_status` and `match_response` (response expression)?

src/Http/ResponseMatcher/CacheableResponseMatcher.php
ddeboer: Yes, in `Response::isCacheable()`. However, Symfony’s list is incomplete, and I wanted to stick to RFC 7234 for good default behaviour.

dbu:
>     i think instead of having additional cacheable states globally or
>     specifically, we should document which services to replace when you
>     want to change the rules.
> 
> That’s fine with me. What about ditching |ResponseMatcher| (that takes
> an expression or status code) and /only/ having
> |NonErrorResponseMatcher| (invalidates) and |CacheableResponseMatcher|
> (set caching headers/tags)? Would people still need the
> |additional_cacheable_status| and |match_response| (response expression)?

hm. okay maybe it would still make sense to have the status list and
expression as a global option.

one each for cacheable and one for must invalidate. i would for example
want to always invalidate cache, even on a 500 response (because part of
the operation might have succeeded and beeing too agressive with
invalidation is less risky than not agressive enough). if there is a
globa configuration for the status codes and expression, its easy to
adjust and could cover some cases.

i feel that there are use cases for this without needing a completely
custom matcher. (i know, i just say the opposite of what i said half an
hour ago...)

ddeboer:
No problem. :wink:

So, I’ll

1. add (global) bundle configuration params for the cacheable and must_invalidate rule matchers;
2. remove the (specific) rule response matching: we only keep the request part.

dbu:
i agree. specific behaviour can still be done on individual controller actions.

src/EventListener/TagListener.php
ddeboer: This cannot be moved to earlier in the method because `$this->symfonyResponseTagger->tagSymfonyResponse($response);` still has to happen: e.g. for when tags are set from Twig. The `testTwigExtension` test fails if we add `if (count($tags)) { return; }` before this.

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -131,6 +134,22 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $this->getAlias().'.cacheable.response.additional_status',
+            $config['response']['additional_status']
+        );
+
+        $definition = $container->getDefinition($this->getAlias().'.response_matcher.cacheable');
+
+        // Change CacheableResponseMatcher to ExpressionResponseMatcher
+        if ($config['response']['expression']) {
+            $definition->setClass(ExpressionResponseMatcher::class)
+                ->setArguments([$config['response']['expression']]);
```
ddeboer: This is a bit odd, but it works: when an expression is configured, switch to another RuleMatcher, which is dedicated to expressions.

ddeboer:
Ready for final review. 

Decided not do document how to do fully custom cacheable/must_invalidate rules (by overriding services) for now. Let’s find out if users really need that before we bloat the docs. :wink:

Resources/doc/reference/glossary.rst
```diff
@@ -5,9 +5,7 @@ Glossary
 
     Cacheable
         A *response* is considered cacheable when the status code is one of
-        200, 203, 300, 301, 302, 404, 410. This range of status codes can be
-        extended with :ref:`additional_cacheable_status` or overridden with
-        :ref:`match_response`.
+        200, 203, 300, 301, 302, 404, 410.
```
dbu: erm, this does not match the list in other places. afaik its the same list that includes 414 and 501, no?

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -131,6 +134,22 @@ public function load(array $configs, ContainerBuilder $container)
         }
     }
 
+    private function loadCacheable(ContainerBuilder $container, array $config)
+    {
+        $container->setParameter(
+            $this->getAlias().'.cacheable.response.additional_status',
+            $config['response']['additional_status']
```
dbu: should we do that in the else if the expression is empty? or do we want this parameter to always be defined? or we could make it not a container parameter but set the argument, like we do for expression.

src/EventListener/TagListener.php
dbu: sorry, you are right.

src/Http/RuleMatcher.php
```diff
@@ -30,66 +30,35 @@ class RuleMatcher implements RuleMatcherInterface
...
-
-    /**
-     * @param RequestMatcherInterface $requestMatcher Strategy to match the request
-     * @param array                   $criteria       Additional criteria not covered by request matcher
-     */
-    public function __construct(RequestMatcherInterface $requestMatcher, array $criteria)
-    {
+    public function __construct(
+        RequestMatcherInterface $requestMatcher = null,
+        ResponseMatcherInterface $responseMatcher = null
```
dbu: can you please update the class doc here to reflect this change?

src/Http/RuleMatcher.php
```diff
@@ -30,66 +30,35 @@ class RuleMatcher implements RuleMatcherInterface
...
      */
-    private $expressionLanguage;
-
-    /**
-     * @param RequestMatcherInterface $requestMatcher Strategy to match the request
-     * @param array                   $criteria       Additional criteria not covered by request matcher
-     */
-    public function __construct(RequestMatcherInterface $requestMatcher, array $criteria)
-    {
+    public function __construct(
```
dbu: please add a method doc that explains that both matchers must match for the rule matcher to match.

and i wonder if RuleMatcher is still the right name for this. its actually a RequestResponseMatcher or a HttpMatcher. not happy with both those names, but well... and we do a BC break here anyways, might as well rename it if we find a better name.

CHANGELOG.md
```diff
@@ -32,6 +32,13 @@ Changelog
 
 * Deprecated methods have been removed.
 
+### Rule matcher
+
+* **BC break:** The `match_response` and `additional_cacheable_status` 
+  configuration parameters were removed for individual match rules. 
+  
+* Cacheable status codes are now configured globally (`cacheable.response`).
```
dbu: we should also mention that the signature of RuleMatcher changed (or that its replaced by RequestResponseMapper or whatever if we decide to rename it)

src/Http/RuleMatcher.php
ddeboer: Done.

ddeboer:
Thanks for the feedback! Updated.

dbu:
great job, thanks a lot!

PR #355, opened 2017-02-10
clarify documentation of cacheable response status

dbu:
follow up of #354

reading through the changes, i realized that expression leads to default status codes being ignored. i think that is fine, but wanted to make the documentation clear about that.


ddeboer:
👍

PR #356, opened 2017-02-10
Refactor UserContextListener to take options array

### LINKED ISSUE #352: refactor UserContextListener to use options array
and use the OptionsResolver for defaults and validation.

improve class doc to distinguish better from the library listener which is within the caching proxy

ddeboer:
Fix #352.

### DIFF
--- a/src/DependencyInjection/FOSHttpCacheExtension.php
+++ b/src/DependencyInjection/FOSHttpCacheExtension.php
@@ -193,12 +193,14 @@ class FOSHttpCacheExtension extends Extension
             ->replaceArgument(0, $config['match']['accept'])
             ->replaceArgument(1, $config['match']['method']);
 
+        $container->setParameter($this->getAlias().'.event_listener.user_context.options', [
+            'user_identifier_headers' => $config['user_identifier_headers'],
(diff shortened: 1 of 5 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -193,12 +193,14 @@ private function loadUserContext(ContainerBuilder $container, XmlFileLoader $loa
             ->replaceArgument(0, $config['match']['accept'])
             ->replaceArgument(1, $config['match']['method']);
 
+        $container->setParameter($this->getAlias().'.event_listener.user_context.options', [
+            'user_identifier_headers' => $config['user_identifier_headers'],
+            'user_hash_header' => $config['user_hash_header'],
+            'ttl' => $config['hash_cache_ttl'],
+            'add_vary_on_hash' => $config['always_vary_on_context_hash'],
```
dbu: i just notice that this means we never actually use the defaults defined in the options, but the ones from the Configuration class. we could change the configuration to NULL, and built the options array to only hold those values that are not null... but this is independent of this refactoring.

dbu:
thank you!

Issue #357, opened 2017-02-13
Questioning the user context


Toflar:
Hey guys,

I was wondering how much v2 is a thing already :-)
I did not know about this bundle, nor the library up until today and actually only found it by chance because I was looking for a solution to implement that "preflight request idea".
By "preflight request idea" I mean the extra request on `/_fos_user_context_hash` to then replicate the `X-User-Context-Hash` onto the real request.
It's a great concept, I just wonder why you are limiting it to user context hashing.
Let me tell you about my example: I have some legacy application that does provide two different page layouts, a desktop one and a mobile one. By setting a cookie, you can force one of the two variants so you already see the problem: I need to `Vary` on something like `Page-Layout: mobile` and `Page-Layout: desktop`.

All the user context listener does in my opinion is extracting information out of the `Cookie` (and/or `Authorization`) header and splitting it up in a `X-User-Context-Hash` header.

Is there any reason why you limited this? I mean, couldn't we have it way more generic?
I'll adjust the docs and rephrase it so you understand what I mean:

1. A client requests `/foo.php` (the original request).
2. The proxy server receives the request. If the request might contain user specific information (`Cookie` or `Authorization` headers) it sends a request (the preflight request) with a special accept header (`application/vnd.fos.replay_headers`) to a specific URL, e.g., `/_fos_replay_headers`.
3. The application receives the preflight request. The application knows the user specific information based on e.g. the `Cookie` header that maps to e.g. a PHP session and thus about things like user context (roles, permissions, etc.) and adds headers based on that information. The application then returns a response containing these headers plus the general header `FOS-Replay-Headers` that contains a list of all headers that should be replayed onto the original request (e.g. `FOS-Replay-Headers: User-Context-Hash, Page-Layout`).
4. The proxy server receives the preflight response, copies all headers that were listed in `FOS-Replay-Headers` onto the client’s original request for `/foo.php` and restarts that request.
5. If the response to this request should differ e.g. per user context, the application specifies so by setting a `Vary: X-User-Context-Hash` header. The appropriate user role dependent representation of `/foo.php` will then be returned to the client.

This would make the whole replay logic reusable for any other use case.
WDYT?

ddeboer:
Hey, thanks for your input!

> I was wondering how much v2 is a thing already :-)

[...]

PR #358, opened 2017-02-15
[RFC] Separate user context from general header replay feature

Toflar:
This is a PR for the discussion in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/357.

Thoughts:

* I did not think of BC at all yet because I think it would make the discussion harder. I want to focus on architecture first.
* I haven't tested a thing yet, I think for discussion purposes this PR should be enough. I'll add full coverage once we agreed on a solution.
* I might have removed some functionality in the `UserContextListener` because I did not understand it. I have to admit that some code in there did not make sense to me :-) Probably because replaying headers and setting the context hash was mixed up, so yet another reason why this should be separated.
* The new `ReplayHeadersEvent` still supports setting the TTL. For this, it always selects the lowest value by any listener set so it works for all of them :-)

Questions:

* Any specific reason why the listener is set to priority 7? Should be commented if so maybe :-)

ToDo's:

* [ ] Tests
* [ ] Replace the `UserContextListener` in `FOSHttpCache` by a general `ReplayHeadersListener` for Symfony's `HTTPCache` support


CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ Changelog
 
 ### Event listeners
 
+* **BC break:**  The preflight request to extract context information from
+  `Authorization` and `Cookie` headers is no longer part of the `UserContextListener`
+  but has instead been moved to a dedicated `ReplyHeadersListener`.
```
stof: wrong class name.

src/Event/ReplayHeadersEvent.php
```diff
@@ -0,0 +1,102 @@
...
+use Symfony\Component\HttpFoundation\ResponseHeaderBag;
+
+/**
+ * Class ReplayHeadersEvent
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class ReplayHeadersEvent extends Event
+{
+    const EVENT_NAME = 'fos.replay_headers';
```
stof: the event name should start with ``fos_http_cache`` instead. We have more bundles in FOS than this one. The bundle config prefix is not just ``fos``

src/EventListener/UserContextListener.php
```diff
@@ -211,7 +136,7 @@ public function onKernelResponse(FilterResponseEvent $event)
     public static function getSubscribedEvents()
     {
         return [
-            KernelEvents::RESPONSE => 'onKernelResponse',
```
stof: this is wrong. You replaced the wrong event

src/Resources/config/replay_headers.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="fos_http_cache.event_listener.replay_headers" class="FOS\HttpCacheBundle\EventListener\ReplayHeadersListener">
+            <argument type="service" id="event_dispatcher" />
+            <argument /> <!-- Identifying headers array -->
```
stof: should be ``<argument type="collection" />``

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -1,47 +0,0 @@
...
-
-namespace FOS\HttpCacheBundle\UserContext;
-
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\RequestMatcherInterface;
-
-/**
- * Matches anonymous requests using a list of identification headers.
- */
-class AnonymousRequestMatcher implements RequestMatcherInterface
```
stof: Removing this class is a BC break (and we may still need it)

src/EventListener/UserContextListener.php
```diff
@@ -188,18 +125,6 @@ public function onKernelResponse(FilterResponseEvent $event)
             ) {
                 $vary[] = $this->options['user_hash_header'];
             }
-        } elseif ($this->options['add_vary_on_hash']) {
```
stof: Why are you removing features from the response event ?

src/EventListener/UserContextListener.php
```diff
@@ -188,18 +125,6 @@ public function onKernelResponse(FilterResponseEvent $event)
             ) {
                 $vary[] = $this->options['user_hash_header'];
```
stof: All the Varying logic probably belongs to the Replay listener instead.

stof:
Not thinking about BC is a mistake IMO. We **must** have a BC layer for the feature, to allow people to migrate without their app without knowing it (if you just replace the feature, people upgrading the bundle without migrating their Varnish stack will start sharing the same cache for everyone, which is very bad, and not clear about what is wrong).
The new architecture must be thought in a way allowing to build a BC layer (which might involve rejecting some solutions considered valid otherwise).

And this also requires checking how this impacts the Varnish configuration in FOSHttpCache, as the discussion in the issue had some concerns about the feasibility of the feature in VCL.
There is no need to spend time on implementing this if it cannot be supported in Varnish (support only in HttpKernel's HttpCache is a no-go, as it is not suitable for cases needing a very good performance, which are the case benefiting the most from a finer grained varying)

Toflar:
Yeah to be honest, I don't see a way BC can be kept. I would never go for `application/vnd.fos.user-context-hash` because it would not correctly represent what it does anymore. That's why I initially asked how much v2 is a thing already because if people migrate to v2, the should also verify their VCL and clear the full cache :-)

My problem about this really is that I don't need the user context at all for my use case. I need a way to replay headers which is one of the many features this bundle and the associated library `FOSHttpCache` provides. But this feature is tightly coupled to the `UserContextListener` which is wrong in my opinion. Moreover, to support replaying headers with the SF HttpCache, I need an `EventDispatchingHttpCache` just like this bundle does. And probably others need those events too.

My alternative would be to

* require the bundle
* disable the user context listener
* replicate half of the `UserContextListener` for my own purposes
* create a custom subscriber for the SF HttpCache support

I mean, I can do that. It's just a pity somewhat but also I understand and I support the fact that you're trying not to change too much.
Maybe I should go for my own solution and we give splitting responsibilities another chance for v3?

stof:
The UserContextListener does not require at all that the logic generating the hash is based on the active user at all.
AFAICT from the discussion on the issue, the *only* thing which is not supported for your use case is to have several independant hashes, to vary different requests separately. And this is precisely the part of the feature for which the feasibility in Varnish has been questioned, meaning it is the first thing to investigate.

Removing ``user`` from the naming of the feature in case the context hash is not based on the user but on something else is just a naming thing.
btw, ``replay_headers`` is much less understandable than ``context`` to me. This feature is about doing contextual caching. I still don't understand what the ``replay_headers`` is meant to represent (except maybe an implementation details of how contextual caching is implemented in the communication between the app and the reverse proxy, but even ``replay`` is not clear for me in this approach of the meaning of the name)

Toflar:
I see your point. And I agree that the name "user context" is not bad at all because `Cookie` and `Authorization` represent a user context so that might be fine. I mean, my example of adding `Page-Layout` as a header could very well be integrated into the context hash. It could just generate useless cache entries. The more context I put into the hash the smaller the chance for a cache hit, right? :-) But for completely individual resources I don't need the proxy anyway, I'd use hinclude if I wanted to optimize that.

Hmm, hmm...so probably I don't need multiple headers at all.

stof:
@Toflar your need is about having multiple independent hashes indeed, to maximize hits. But then, start by evaluating whether this is achievable in Varnish.

dbu:
@stof for context, see #357

one idea was to document how to write custom vcl if you need anything else than user context. i am optimistic that we can generalize this while keeping BC. the replay header for user context can stay the same. if that is the only header, there is no need to change the varnish config, and you can ignore the header spec. if you have additional headers, you can either do custom vcl or use some inline C hackery to make thing generic.

on symfony side, we can make it BC on the level of configuration and by keeping the old classes around, but changing the service definitions to register the user context as one of the things in replay headers.


@Toflar i need to find time to thorougly review the pull request. i fear this will only happen next week. i am very busy at the moment.

Toflar:
> @Toflar your need is about having multiple independent hashes indeed, to maximize hits. But then, start by evaluating whether this is achievable in Varnish.

@stof Of course it should be possible. I mean, I did not test it but you just have to add the other headers too. You cannot automate it like we could for the SF reverse proxy as this would apparently require some inline C or custom varnish module but you can of course manually configure all the headers. For illustration purposes:

```diff
@@ -4,6 +4,7 @@ ub vcl_recv {
     if (req.restarts == 0
         && (req.http.accept ~ "application/vnd.fos.user-context-hash"
             || req.http.X-User-Context-Hash
+            || req.http.Page-Layout
         )
     ) {
         return (synth(400));
@@ -67,6 +68,7 @@ sub vcl_deliver {
         && resp.http.content-type ~ "application/vnd.fos.user-context-hash"
     ) {
         set req.http.X-User-Context-Hash = resp.http.X-User-Context-Hash;
+        set req.http.PageLayout = resp.http.Page-Layout;
 
         return (restart);
     }
@@ -76,6 +78,7 @@ sub vcl_deliver {
     # Remove the vary on context user hash, this is nothing public. Keep all
     # other vary headers.
     set resp.http.Vary = regsub(resp.http.Vary, "(?i),? *X-User-Context-Hash *", "");
+    set resp.http.Vary = regsub(resp.http.Vary, "(?i),? *Page-Layout *", "");
     set resp.http.Vary = regsub(resp.http.Vary, "^, *", "");
     if (resp.http.Vary == "") {
         unset resp.http.Vary;
@@ -83,4 +86,5 @@ sub vcl_deliver {
 
     # Sanity check to prevent ever exposing the hash to a client.
     unset resp.http.X-User-Context-Hash;
+    unset resp.http.Page-Layout;
 }
```

> @Toflar i need to find time to thorougly review the pull request. i fear this will only happen next week. i am very busy at the moment.

@dbu no problem at all. I might still work on it, depending on @stof's feedback ;-)

@stof I understand you don't like "replay headers". We might also stick with user context, like so:

* UserContextListener
    * HashListener
        * Hash provider A
        * Hash provider B
    * PageLayoutListener

The only thing that's not so nice is that for BC reasons you might want to keep `Content-Type: application/vnd.fos.user-context-hash` altough `Content-Type: application/vnd.fos.user-context` would be more precise because it will no longer only work with the hash.
And this only affects BC of the proxy configuration. The listener itself wouldn't be BC of course but there are BC breaks documented already so I don't quite understand why those were allowed but I'm not allowed to add more :-D

dbu:
for BC, i think we can act on the Accept header. if its the old one, we return with the old content-type and throw an exception if more than the user context is configured. if its the new accept type, we do the new logic.

Toflar:
Yeah we could also just leave the `UserContextListener` completely as is, deprecate it and build a new one with a different accept header (by default). I just wonder how we should name it then 🤔

ddeboer:
To summarise, both replaying and user context are actually about *collecting clients into groups* based on the Cookie and Authorization request headers and serving these groups *response variants*:
- UserContextListener was originally meant to group clients *as users* into specific roles (e.g. anonymous, member, editor). The advantage is that [user context providers](http://foshttpcache.readthedocs.io/en/stable/user-context.html#context-providers) can be changed without the need for changing the proxy server config (VCL), because all aspects are combined into a single header.
- ReplayHeadersListener intends to group clients based on some other properties (e.g. a display preference setting). It *replays* response separate headers from the preflight response on the actual request. The advantage is more efficient caching because of more fine-grained variants: one response varies on user role; another on user preference.

Given these differences, I think we should support both approaches. 

1. The new Accept header name becomes `application/vnd.fos.user-context` (instead of `application/vnd.fos.user-context-hash`). 
2. We rename UserContextListener to UserContextHashListener, which handles the old Accept header. The UserContextProviders’ values are hashed and combined into one header.
3. We introduce a new UserContextListener that handles the new header. The UserContextProviders’ values are stored in separate headers.

Toflar:
@ddeboer perfect summary, you got it all right. There's only one issue with this: the reverse proxy configuration needs to know 3 things:

* The `Content-Type` to listen to (default: `application/vnd.fos.user-context-hash`)
* The "preflight request endpoint" or path (default: `_fos_user_context_hash`)
* The headers to take from the preflight request response to the original (replaying, replicating, whatever) (default: `X-User-Context-Hash`)

**All of which** is needed for the new feature as well. I understand that you try to keep BC but we're talking about version 2 of this bundle and library and already introduced BC breaks so I just don't understand why we cannot introduce more. If we leave the current listener as it is, I am very sure it will cause confusion for the users:

* Why are there two content types `application/vnd.fos.user-context` and `application/vnd.fos.user-context-hash`? Also, I can configure both. Must they differ and if so, why? 
* Why is the path named `_fos_user_context_hash`, if I don't need the hash feature?
* If I need the user context hash feature and I need to replay my `Super-Header`, do I need to have more than one preflight request or is `_fos_user_context_hash` also handling my `Super-Header` even though the path does not say so? Also, do I need to handle both `application/vnd.fos.user-context` and `application/vnd.fos.user-context-hash` in my Varnish configuration?

If we'd introduce another BC break by defaulting to `application/vnd.fos.user-context` and `_fos_user_context` for **both**, this would not happen. There would be only one `Content-Type` that needs to be handled by your Varnish configuration. There would be only one path for the preflight request. If you've already used this bundle in v1, all you need to do when upgrading to v2, is replacing all occurrences of `application/vnd.fos.user-context-hash` by `application/vnd.fos.user-context` and `_fos_user_context_hash` by `_fos_user_context` in your Varnish configuration. Done. That's all you need to do if you use the defaults.
If you want to replay your `Super-Header` now because that's a neat new feature, you can read up on the docs for it but basically it is as simple as searching for `X-User-Context-Hash` and adjust the VCL so that it also handles `Super-Header`.

Toflar:
So guys. How do we proceed here?

dbu:
i think we should rename and refactor the service, but in a way that it still can handle the old accept type with the simple hash behaviour. that will keep working with varnish without special modules or whatever will be needed. that probably means that along with the generic list of providers, we inject the user context hash provider separately (a second time) and then the listener can use the header to chose its mode.

we can add the second route as well. the routes do not really matter that much, as the listener happens before any controller is called. but the route must exist, or the listener is never called because it happens after routing, when security is loaded. maybe we just want to use / as route, instead of anything specific. we have the content type to distinguish what we want to do.

Toflar:
Okay, I'll see what I can do.

dbu:
thanks you. and sorry for the silence. so many things to do and so little time... at list my stand on BC for this is: caching proxy communication must be BC over one version, as otherwise you would need to perfectly sync varnish and symfony deployments which can be cumbersome. but php wise, class names and configuration may change as we see fit (as long as the new configuration is an improvement, of course. i do think your proposal does add value - challenge will be to find something that is not too complicated).

ddeboer:
If we want both BC and a clean slate, let’s do it like this:

1. @Toflar adds a new listener that acts on the new values (content type `application/vnd.fos.user-context`, endpoint `_fos_user_context` and header `X-User-Context`). We can call this listener `UserContextListener`.
2. We rename our current `UserContextListener` to `LegacyUserContextListener` and add deprecation warnings to it. For now, it continues to work on the previous content type and header name. Requests to the old endpoint `_fos_user_context_hash` or with the old header `X-User-Context-Hash` will trigger a warning, however.
3. When enabling the user context in the bundle config, both listeners are activated. As long as you use the new config, `LegacyUserContextListener` won’t be triggered and no deprecation warnings are shown.
4. We update the docs for the new setup and explain how to upgrade to it from the previous `application/vnd.fos.user-context-hash` etc. Like @Toflar mentioned: 

    > replacing all occurrences of application/vnd.fos.user-context-hash by application/vnd.fos.user-context and _fos_user_context_hash by _fos_user_context

dbu:
i am not sure we can have a really clean slate. varnish won't be able to
handle generic headers without inline C or some plugins.

i think we need to keep support for the current user-context-hash
concept. so i would not rename the current listener to legacy, and
phpdoc the two variants.

@Toflar do you want to work through this and refactor for the two
options, and provide the doc? i think its a considerable effort to make
this not confusing and understandable.

Toflar:
Yeah to be honest, this discussion sort of makes me think about building my own bundle for that and just use the `FOSHttpCacheBundle` as a dependency. Don't get me wrong, I think the discussion was extremely important because it showed that integrating it into the core of `FOSHttpCacheBundle` isn't as simple as it might look at first sight. So I'd like to thank everybody that participated. I think a big part of why we're unsure about how to integrate is due to the fact that we're talking about a new feature that does not exist yet. I think if I build my own bundle first, we'll have something that really exists in the wild (maybe gets some attention and adjustments) and we can talk about integrating maybe in `FOSHttpCacheBundle` v3 again. 

WDYT?

dbu:
i think this is a good way to move forward. i expect the bundle will be really simple with just this specific task, making it easy to understand and we can see how well that works (and also if other people have similar use cases and if their cases are covered by it). and then integrate things back into this bundle once we see things more clearly.

thanks for the discussion @Toflar, it was indeed very interesting and valuable.

Toflar:
I'll close this once my bundle is published :) May I ask another question?

Why do you require the very same listener (more or less) to be implemented twice when you want to have support for the Symfony HttpCache? If I look at the `EventDispatchingHttpCache` trait, it works like this:

```php
public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
{
    if ($response = $this->dispatch(Events::PRE_HANDLE, $request)) {  // Here, the symfony cache user context listener is more or less duplicated because it adds the hash stuff.
        return $this->dispatch(Events::POST_HANDLE, $request, $response);
    }

    $response = parent::handle($request, $type, $catch); // Regular request where the regular user context listener must be ignored

    return $this->dispatch(Events::POST_HANDLE, $request, $response);
}
```

This requires 2 times the same listener. Why not like this:

```php
public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
{
    if ($response = $this->dispatch(Events::PRE_HANDLE, $request)) {  // Nothing is done here
        return $this->dispatch(Events::POST_HANDLE, $request, $response);
    }

    $response = parent::handle($request, $type, $catch); // Regular request where the regular user context listener kicks in and adds his hash headers

    return $this->dispatch(Events::POST_HANDLE, $request, $response); // Allow to call parent::handle() again for the real request.
}
```

That way there's no need to register two events and duplicate the code. If this was done for performance reasons, I don't really understand because then you switch to Varnish anyway :-)

Can I do a PR for this?

Toflar:
Ah, I noticed there's a `getKernel()` on the event so I can do this on my own maybe :)

dbu:
i can't find the difference between the two code snippets you posted. where is the other coming from? (note that version 1 of the bundle and library are quite a bit different there from version 2, because we went to a php version that allows traits)

Toflar:
Oh, the difference is just my comments :) So when you are using Varnish, you need a regular `kernel.request` event listener that handles the user context, right? So that all makes sense, you need a request listener to do that. When you are using SymfonyCache, however, you seem to use the `PRE_HANDLE` event of the `EventDispatchingHttpCache` to do the almost very same in a different listener again (in the `FOS\HttpCache\SymfonyCache\ UserContextListener`). Although you could just let the original request pass trough so the same event listener that's used for Varnish too kicks in and sends an early response and then use the `POST_HANDLE`event to check the result of the event listener and send the request with the hash header to the application again.

dbu:
ah no, [that listener](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/UserContextListener.php) is the other part of user context: it handles the caching proxy part of user context, not the web application part.

Toflar:
I've just published the bundle: https://github.com/terminal42/header-replay-bundle
Feel free to leave feedback there, now I just have to wait for the FOS bundle and lib to be published as v2 stable :) Are there any plans on this?
Also, feel free to close this PR and the related issue I opened.

dbu:
great job! i read over the README and think it explains things quite well. one input i have would be to write a short paragraph above the why that explains the goal of the bundle, without assuming that people know what a preflight request is.

i will close this pull request now. but please do a pull request against our documentation to add a link to your bundle in the place where we talk about user context.

Toflar:
Thx for the feedback, very much appreciated! I extended the docs with a diagram which I hope will support the readers in understanding the concept. PR for the docs is here: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/363.

Any comment on my release question? 😄

dbu:
> Any comment on my release question? 😄

i did tag 2.0.0-beta1 the other day. i think we are pretty close to
release actually.

PR #366, opened 2017-05-25
Adjust to tag header cleanup and allow more configuration on varnish proxy client

### LINKED ISSUE #281: [RFC] Do not force client for TagHandler or other one
When using the TagHandler user have to set a proxy client to do the invalidation. 

However we got an use case where we have multiple applications, and one is doing the invalidation and the other only set the tags and we are forced to set a client in the second one but it's useless in this case.

I don't know if the current architecture of TagHandler allow to only add tag on response without having a Client ? If it's the case, can this be separated in another class and remove the dependency ?

dbu:
fix #281 

/cc @Toflar


composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^5.6.0||^7.0.0",
-        "friendsofsymfony/http-cache": "^2.0.0-beta1",
+        "friendsofsymfony/http-cache": "2.0.x-dev as 2.0.0-beta3",
```
dbu: change to `^2.0.0-beta3` when released

dbu:
yay, tests are green now.

@Toflar would you mind checking if this looks good for you and works? then i will tag new RC of both library and bundle.

composer.json
dbu: ah snap. will fix this in master :-/

dbu:
thanks a lot for the refactoring of the library and the review here @Toflar . i just tagged 2.0.0-beta2 of the bundle. this hopefully can be the last beta.

PR #372, opened 2017-06-20
cache control listener should only set caching headers on cacheable responses

dbu:
the documentation on http://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/cacheable.html claims that cache-control headers are only set on cacheable responses. but we missed to inject the cacheable matcher to the cache control listener.

this is a BC break for the listener class, but i think the bug is severe enough that we want to fix it nonetheless. the listener is more of an internal class to implement the bundle behaviour. should we mark the listeners as `@internal` and say you are not supposed to code-reuse them?

Options
* [ ] If we finalize this approach, adjust the unit test CacheControlListenerTest to pass a rule matcher to the CacheControlListener and write another functional test.
* [ ] Or we could change addRule of the CacheControlListener back to RuleMatcher instead of RequestMatcher and re-introduce match_response configuration. However, this leads to weird interactions with the global cacheable configuration. 

I opt for a note in the documentation to say that if you want headers to be always set regardless of response status, you should write your own event listener.

FTR: we removed the response matching in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/354 but from how i read this we mostly were thinking about invalidation and tagging and less about the header rules.

### DIFF
--- a/src/EventListener/CacheControlListener.php
+++ b/src/EventListener/CacheControlListener.php
@@ -11,6 +11,7 @@
 
 namespace FOS\HttpCacheBundle\EventListener;
 
+use FOS\HttpCacheBundle\Http\RuleMatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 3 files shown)


src/EventListener/CacheControlListener.php
```diff
@@ -55,10 +64,12 @@ class CacheControlListener extends AbstractRuleListener implements EventSubscrib
     private $debugHeader;
 
     /**
-     * @param string|bool $debugHeader Header to set to trigger debugging, or false to send no header
+     * @param RuleMatcherInterface $cacheableMatcher Matcher to decide whether caching headers should be set
+     * @param string|bool          $debugHeader      Header to set to trigger debugging, or false to send no header
      */
-    public function __construct($debugHeader = false)
+    public function __construct(RuleMatcherInterface $cacheableMatcher, $debugHeader = false)
```
dbu: we could avoid the BC break by using = null for the matcher and create a default one if none is passed. would that be the right thing to do or too many options?

dbu:
closing in favor of #373

PR #373, opened 2017-06-20
cache control listener needs to also look at response

dbu:
alternative to #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,21 +1,29 @@
 Changelog
 =========
 
-2.0.1
+2.1.0
 -----
 
 ### Added
 
 * Individual rules in the `cache_control` can now again have a `match_response`
-  or `additional_response_status` configuration to limit the rule to certain responses. 
-  If not configured, the global `cacheable` configuration is now respected.
-  **BC BREAK**: The signature of CacheControlListener::addRule has changed. It
-  now expects a RuleMatcherInterface instead of the ResponseMatcherInterface.
-  **BC BREAK**: The global `cacheable` configuration is now respected when
-  adding cache headers. By default, this follows RFC 7234, only responses with
-  status 200, 203, 204, 206, 300,  301, 404, 405, 410, 414 or 501 get cache
-  headers. You can change this list in `cacheable` or with the re-added
-  `match_response` / `additional_response_status` for individual rules.
+  or `additional_response_status` configuration to limit the rule to certain
+  responses.
+
+  For this, the signature of CacheControlListener::addRule had to be changed.
+  It now expects a RuleMatcherInterface instead of the
+  ResponseMatcherInterface. If you extended the listener or change the service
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+2.0.1
+-----
+
+### Added
+
+* Individual rules in the `cache_control` can now again have a `match_response`
+  or `additional_status` configuration to limit the rule to certain responses.
```
dbu: should we call this additional_response_status to be clear? on cacheable, its response.additional_status

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -186,6 +213,37 @@ private function parseRuleMatcher(ContainerBuilder $container, array $match)
         );
     }
 
+    /**
+     * Used only for cache control rules.
+     *
+     * @param ContainerBuilder $container
+     * @param array            $config
+     */
+    private function parseResponseMatcher(ContainerBuilder $container, array $config)
```
Tobion: missing `@return`

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -174,6 +174,33 @@ private function loadCacheControl(ContainerBuilder $container, array $config)
 
     private function parseRuleMatcher(ContainerBuilder $container, array $match)
     {
+        $requestMatcher = $this->parseRequestMatcher($container, $match);
+        $responseMatcher = $this->parseResponseMatcher($container, $match);
+
+        $signature = serialize([(string) $requestMatcher, (string) $responseMatcher]);
+        $id = $this->getAlias().'.cache_control.rule_matcher.'.md5($signature);
+
+        if (!$container->hasDefinition($id)) {
```
Tobion: I don't see the point of this. This can't be defned. And if it was there is an error

dbu: the service ids of request + response matchers are based on the md5 of the rules in them. so if both are exactly the same as a previous rule, it could happen. 

hm, ok. while for response or request matcher alone that case makes sense, for a rule matcher it makes no sense as they are only used for cache_control and are the full definition of when a rule matches, hence a duplicate indicates a rule that is always hidden by a previous one.

should we instead throw an exception if the id already exists and complain that this rule is hidden by a previous one? could be helpful for the user to spot misconfiguration

dbu:
@Tobion thanks for the feedback. changed duplicate rules to configuration exceptions. does this look sane to you now?

CHANGELOG.md
```diff
@@ -1,6 +1,30 @@
 Changelog
 =========
 
+2.0.1
```
Tobion: Are you sure this is released as 2.0.1 and not as 2.1 considering it added features?

dbu: agreed, thanks

PR #380, opened 2017-08-31
Fix deprecated console.exception event

Tobion:
This event is deprecated since symfony 3.3. And listening to it seems useless here anyway because the terminate event is already used which is triggered after the exception event in any case. So they are not exclusive.


dbu:
thanks. that was probably a misunderstanding when we added this.

PR #381, opened 2017-08-31
test that services can actually be loaded

### LINKED ISSUE #334: functional test if services can be created
noted in #333 it seems that some services are never loaded in the tests, which can hide problems like the one fixed there.

dbu:
fix #334

### DIFF
--- a/tests/Functional/Fixtures/app/config/config.yml
+++ b/tests/Functional/Fixtures/app/config/config.yml
@@ -47,7 +47,7 @@ fos_http_cache:
   test:
     proxy_server:
       varnish:
-        config_file: varnish.vcl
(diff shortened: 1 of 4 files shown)


tests/Functional/Fixtures/app/config/config.yml
```diff
@@ -47,7 +47,7 @@ fos_http_cache:
   test:
     proxy_server:
       varnish:
-        config_file: varnish.vcl
+        config_file: %kernel.root_dir%/../../../../vendor/friendsofsymfony/http-cache/tests/Functional/Fixtures/varnish/fos.vcl
```
dbu: needs to be a real file now.

PR #382, opened 2017-08-31
testing that user context request is handled correctly

dbu:
* [x] figure out how to handle the problem that the request by the test client is not a MASTER_REQUEST.


Toflar:
For the record: This is because the route

```yaml
user_context_hash:
    path: /_fos_user_context_hash
```

is missing in the tests.

PR #388, opened 2017-09-28
test with php 7.2

dbu:


dbu:
this is quite puzzling. whats more, it looks like the test assumption was expecting a wrong behaviour and php 7.2 magically fixes this. the code in question is https://3v4l.org/IOvBR - and it makes no sense to me why the result is different on 7.2.

dbu:
looks like we end up with the wrong version of phpunit. maybe we should get away from the symfony phpunit bridge and directly depend on phpunit again?

Issue #389, opened 2017-09-29
FlashMessageListener constructor is invalid


wesnick:
In this commit:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/commit/1527e44fb6c37a66573bae3e90f133984ba5eac8 , OptionsResolver was introduced.

Before you can set AllowedTypes you must set known options, hence, this class always throws an exception when instantiating.

dbu:
fixed in #391

Issue #34, opened 2014-01-31
[Test] Missing test for FlashMessageListener


dbu:
create a unit test for the flash message listener
- no session started if none yet
- no cookie set if no flash messages
- cookie set when we have flash messages.

PR #390, opened 2017-09-29
Fix FlashMessageListener

wesnick:
- fixes OptionsResolver undefined values exceptions
- fixes Symfony >3.3 flash messages do not start session


wesnick:
Addresses #389 and #34

src/EventListener/FlashMessageListener.php
```diff
@@ -76,13 +77,7 @@ public function onKernelResponse(FilterResponseEvent $event)
...
 
-        // Flash messages are stored in the session. If there is none, there
-        // can't be any flash messages in it. $session->getFlashBag() would
-        // create a session, we need to avoid that.
-        if (!$this->session->isStarted()) {
-            return;
-        }
-
+        // As of Symfony 3.3 directly accessing the flashbag does not start a session.
         $flashBag = $this->session->getFlashBag();
```
wesnick: This change may accidentally start a session for Symfony <3.3

src/EventListener/FlashMessageListener.php
```diff
@@ -48,6 +48,7 @@ public function __construct($session, array $options = [])
         $this->session = $session;
 
         $resolver = new OptionsResolver();
+        $resolver->setDefined(['name', 'path', 'host', 'secure', 'enabled']);
         $resolver->setAllowedTypes('name', 'string');
         $resolver->setAllowedTypes('path', 'string');
         $resolver->setAllowedTypes('host', 'string');
```
dbu: according to the default configuration, the host may also be `null`, actually. although with the code below, i am not sure that this will work out fine. line 91 about `if (isset($cookies[$this->options['host']][$this->options['path']][$this->options['name']])) {` will never work if host is null i think.

if you use the listener, can you check if that check makes sense and works as intended?

dbu:
thanks for the updates. can you revert the change on session started check, so that we for now keep doing it and not start a session on symfony < 3.3? maybe instead add a note that this can be dropped when we drop support for older symfony versions. or even add a kernel version check in the `if` so we save the method call to check whether the session exists, when not needed.

wesnick:
I am trying to run the tests with symfony < 3.3 to see how session behaves differently between the versions.  My composer skills are poor, I guess, because I am struggling to get it to install lower versions of symfony without creating additional errors. The issue is that in Symfony 3.3, session is not started by simply adding a flash message to the flashbag; this is why I removed the session started check.  I will see if I can add an additional test to ensure session is not started accidentally by the listener

dbu:
thanks. the test is the best approach, then we know for sure :+1:

wesnick:
I didn't see that accessing the flashbag starts a session in previous versions of symfony.  I know that doing it in twig with the app helper does start a session, but it does not appear to when interacting directly with the session service.  Let's see what the tests say

tests/Functional/Fixtures/app/config/config.yml
```diff
@@ -48,6 +48,12 @@ fos_http_cache:
     proxy_server:
       varnish:
         config_file: %kernel.root_dir%/Resources/varnish/fos.vcl
+  flash_message:
+    enabled: true
+#    host: mydomain.com
```
dbu: did you mean to commit this line? we should either delete it or keep it active.

dbu:
looking at https://github.com/symfony/symfony/blob/7d85ff27bd421efc32caab0ed3cd6e48d9148de8/src/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php#L277 i think accessing the flash bag would start the session. the tests seem to say differently, but i can't see why... am i missing something? you assert that there is only 1 cookie, so it should be correct - if a session was started, there would be a session cookie... or are the tests maybe not using the NativeSessionStorage but something else, and that something else behaves differently?

wesnick:
looks like the fixture TestSession does not behave as a normal session:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/tests/Functional/Fixtures/Session/TestSessionStorage.php

wesnick:
@dbu I was just experimenting with using MockSession storage, it seems to do what we need, more accurately test existing session.  If this approach seems reasonable to you I can clean up.

dbu:
yep. when i add a `$this->start()` into TestSessionStorage::getBag, the assumption that the session has not started fails. so this is a test behaviour difference. (btw, weird that this is used - its in the tests so not meant to be reused. and there is MockArraySessionStorage with a much cleaner behaviour). hah, just found the reason for that, we instantiate that in the test kernel. that seems wrong, we should either use the mock array thing there, or not overwrite session storage at all.

i quickly tried disabling the overwriting, but then i get an error from expression language that i have no idea if we just never saw it because of the poor mock store, or if its something else.

tests/Functional/Fixtures/Session/TestSessionStorage.php
```diff
@@ -13,111 +13,112 @@
...
+//    public function isStarted()
+//    {
+//        return $this->started;
+//    }
+//
     /**
      * {@inheritdoc}
      */
-    public function getId()
+    public function generateId()
```
dbu: we do we need to overwrite this? do we have an assertion that requires it?

wesnick:
I got a bit more than I bargained for on this issue.  If we replace custom Session implementation with Symfony Mock Session, we see that user context invalidation listener does not actually work as expected.  It turns out that the priority of the logout listeners is not correct, so the InvalidateSessionListener goes first, destroys the session, then the invalidate listener executes, requests a session Id, which will generate a new session Id.  I think the priority of this listener should be changed, but this issue is now creeping a bit.

wesnick:
from the compiled container:
```
        $i = new \Symfony\Component\Security\Http\Firewall\LogoutListener($b, $h, new \Symfony\Component\Security\Http\Logout\DefaultLogoutSuccessHandler($h, '/'), array('csrf_parameter' => '_csrf_token', 'csrf_token_id' => 'logout', 'logout_path' => '/secured_area/logout'));
        $i->addHandler(new \Symfony\Component\Security\Http\Logout\SessionLogoutHandler());
        $i->addHandler(${($_ = isset($this->services['fos_http_cache.user_context.logout_handler']) ? $this->services['fos_http_cache.user_context.logout_handler'] : $this->get('fos_http_cache.user_context.logout_handler')) && false ?: '_'});

```

dbu:
yeah, that goes beyond the scope of this :-/

can you please mark the test as skipped, with a link to your comment above? so that we see that otherwise the tests are green? i will then fix the priorities in a separate PR.

dbu:
thanks a ton! i squashed the commits and merged this in #391 and created #392 for the logout handling problem you found.

i'd like to wait with the release a bit to see if we manage to fix the logout handler as well.

spointecker:
Would be awesome to have this fix available soon! Have been struggling with it today :(

spointecker:
While the official fix is not released I tried the dev-master to be able to work on my pagecache. 
I am having now troubles with the flash messages listener. The flash messages I add in my controller never are recognized within the flash message listener.
to be more explicit: 
`FOS\HttpCacheBundle\EventListener\FlashMessageListener#89 -> $flashes is empty`

Even though the messages appear if I print them in my templates via twig. Can this somehow be related to this or should I raise another issue?

wesnick:
I don't think this change would cause the issue you are seeing.  The listener does not interact with the flashbag until that point in the Listener (line 86).  Keep in mind that if you request the flash messages from the flash bag, the are cleaed from the flashbag.  You might check to see if you are clearing flashes in another spot in your code either by calling ->all() or some other method.  If you must peek at the flash messages in a non destructive way you can call ->peekAll()

spointecker:
Hi @wesnick thanks for your reply!
I think you were right, the printing in TWIG clears the messages. I really thought that the template calls are AFTER the FlashMessageListener and so it would be possible to use the page cache on some pages and on all other I would print flash messages via TWIG.
Isn´t that possible to use TWIG for printing on NOT cached pages and use the JS solution for cached pages? OR is the flash message listener also used on NOT cached pages?

dbu:
the listener does not look at cache headers of the response, so it does not know that. also, twig happens happen before the cache instructions are final, so i think its impossible to generically know. i would recommend to not implement 2 ways of flash messages and stick with the js solution for all messages if you want to use the listener.

note that without further step, the response is not cached when there is a flash message, because of the set-cookie header. you would have to do some more work on varnish to still cache but remove the set-cookie header when serving from cache.

spointecker:
Ok thanks that helps a lot!
I am using the Symfony cache for now (not varnish) - in my experience the flash messages are stored in the cache, if I do not handle that in a kernel response listener. 
So maybe its easier to not store "flash message pages" at all (in my case this shouldnt happen that often)

dbu:
sorry, i was actually confused in my comment above. 

there are two ways of handling a POST: you can reply with a HTML page, or you can redirect to some confirmation page (or the homepage or whatever). 

showing HTML for a POST request has the drawback that reloading that page will re-send the request which sometimes is not desired.

now the thing with caching is that you want to avoid sessions. the POST can not be cached anyways. if you set a flash message, that starts a session for the user (because flash messages are stored in a session). the user will now not get cached results anymore, until you manually terminate the session.
with the flash listener, the message is instead put into a specific cookie for flash message and no session should be started. that flash cookie is only used by some javascript you include on client side, and you set up your AppCache to strip all cookies except session cookei from the request. the response to the POST will send the cookie (and can't be cached anyways, because this was a POST). when the POST redirects, the client will send the cookie, but you remove that. thus you end up with a cache hit for the redirect target. the cookie is at the client side and the js displays the message.

dbu:
i keep being confused by this. will revisit http://foshttpcachebundle.readthedocs.io/en/latest/features/helpers/flash-message.html and write what i just wrote here...

spointecker:
ok I am also getting confused now.
Let´s assume I use the JS option. That means I need to use the page cache on all pages I want to display flash messages, because otherwise the flash messages don´t get stored into the cookie. Is this correct? 
this would be a problem in my case, cause I cant cache all my pages (for now) but I want to be able to print flash messages on every page.

dbu:
i tried to clarify the documentation in #393. can you look at that and see if it becomes clearer, and if your question remains after reading that doc section? if its still unclear, i want to figure out how to explain it in the doc so its understandable...

wesnick:
> with the flash listener, the message is instead put into a specific cookie for flash message and no session should be started. 

Actually, I think a session is started, even though we cleared the flash messages in the FlashMessagesListener, we don't have any easy way to "unstart" the session.  You can see that the test now indicates there are 2 cookies, 1 session cookie and 1 flashes cookie.

spointecker:
It totally makes sense to me. In theory at least. But I keep getting strange results.

My very simple current setup results in a strange behaviour:
- `flash_message: enabled`
- Setting a flash message in an action: `$this->addFlash('success', 'YOOOOO LOOOO ' . time() . rand(1,11111));`
- Caching enabled on that action/page/route
- Using the JS for showing the flash messages from the cookie (from documentation)

So the page cache works, the flash message gets displayed after page load, BUT if I reload the page, I keep getting the same message (see random/timestamp).
The cookie is deleted after the pageload (checked that). I have absolutely no clue where the text of the messages comes.
The cache is HIT on the second pageload.

dbu:
if your twig template renders flash messages, the listener won't help. i think it should not even see the flash messages anymore, because rendering them should remove them. maybe look at the response with the firefox/chrome dev tools to be sure what is in the html and that the flash is only in the cookie, not rendered in html.

> we don't have any easy way to "unstart" the session

hm, thats a problem. any idea if / how we could completely avoid the session from being started? we could unset the set-cookie for the session cookie in the AppCache / varnish, but thats a hack and won't work if some things actually do need a session.

PR #394, opened 2017-10-13
Fix ContextInvalidationLogoutHandler

### LINKED ISSUE #392: Fix ContextInvalidationLogoutHandler
In #390 we replaced the test session handler with the ArrayMockSessionHandler that behaves a lot closer to the real NativeSessionHandler. We now have to skip the ContextInvalidationLogoutHandlerTest because it does not work. The problem is that SecurityBundle SecurityExtension puts `security.logout.handler.session` as the first handler of the session logout handler. The logout handlers seem to not have priorities. That means that the session is already destroid when our context invalidation handler is executed, and it does not have access to the session id anymore.

~There is a 3rd parameter TokenInterface to the LogoutHandlerInterface::logout method. would that be what we need?~ (no, its the username and password thingy) 
we could try to decorate the successhandler, because that is called first.

On a side note: the accept header is not used in BAN by the default ban vcl. should we add handling for that?

wesnick:
This fixes #392 

* [x] doc update
* [x] changelog
* [x] unit test
* [x] deprecate old handler

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,10 @@ Changelog
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
   
-* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
-  ContextInvalidationSessionLogoutHandler.  The original handler was called 
-  after the invalidation of the session, and thus did not invalidate the seesion
-  it should have but a newly created one.  You should remove the deprecated service
+* The `ContextInvalidationLogoutHandler` has been deprecated in favor of the
+  `ContextInvalidationSessionLogoutHandler`. The original handler was called
+  after the invalidation of the session, and thus did not invalidate the session
+  it should have but a newly created one. You should remove the deprecated service
   `fos_http_cache.user_context.logout_handler` from the logout.handlers section
   of your firewall configuration.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,10 @@ Changelog
   
 * The ContextInvalidationLogoutHandler has been deprecated in favor of the 
   ContextInvalidationSessionLogoutHandler.  The original handler was called 
-  after the invalidation of the session so did not correctly invalidate the 
-  session ID in cache.  You should remove the deprecated service 
+  after the invalidation of the session, and thus did not invalidate the seesion
+  it should have but a newly created one.  You should remove the deprecated service
   `fos_http_cache.user_context.logout_handler` from the logout.handlers section
-  of your firewall. 
+  of your firewall configuration.
 
 2.1.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ Changelog
 
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
+  
+* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
+  ContextInvalidationSessionLogoutHandler.  The original handler was called 
+  after the invalidation of the session so did not correctly invalidate the 
+  session ID in cache.  You should remove the deprecated service 
+  `fos_http_cache.user_context.logout_handler` from the logout.handlers section
+  of your firewall. 
 
 2.1.0
 -----
--- a/Resources/doc/reference/configuration/user-context.rst
+++ b/Resources/doc/reference/configuration/user-context.rst
@@ -139,7 +139,7 @@ or you will end up with mixed up caches.
 ~~~~~~~~~~~~~~~~~~
 
 The logout handler will invalidate any cached user hashes when the user logs
-out. This behavior applies to all stateful firewalls.
+out.
 
 For the handler to work:
 
@@ -147,10 +147,11 @@ For the handler to work:
 * Symfony’s default behavior of regenerating the session id when users log in
   and out must be enabled (``invalidate_session``).
 
-Should your application require you to have a per-firewall user context invalidation
-strategy, you should create your own custom context invalidation handler.  Keep in mind that,
-by default, Symfony's LogoutSuccessHandler will place the SessionLogoutHandler first of any
-enabled logout handlers when ``invalidate_session`` is enabled.
+.. note::
+    The logout handler is active on all firewalls.  If your application has multiple firewalls
+    with different user context, you need to create your own custom invalidation handler. Be
+    aware that Symfony's ``LogoutSuccessHandler`` placea the ``SessionLogoutHandler`` before
+    any configured logout handlers.
 
 enabled
 """""""
--- a/Resources/doc/reference/configuration/user-context.rst
+++ b/Resources/doc/reference/configuration/user-context.rst
@@ -150,8 +150,8 @@ For the handler to work:
 .. note::
     The logout handler is active on all firewalls.  If your application has multiple firewalls
     with different user context, you need to create your own custom invalidation handler. Be
-    aware that Symfony's ``LogoutSuccessHandler`` placea the ``SessionLogoutHandler`` before
-    any configured logout handlers.
+    aware that Symfony's `LogoutSuccessHandler` places the `SessionLogoutHandler` before any
+    configured logout handlers.
 
 enabled
 """""""
--- a/Resources/doc/reference/configuration/user-context.rst
+++ b/Resources/doc/reference/configuration/user-context.rst
@@ -139,7 +139,7 @@ or you will end up with mixed up caches.
 ~~~~~~~~~~~~~~~~~~
 
 The logout handler will invalidate any cached user hashes when the user logs
-out.
+out. This behavior applies to all stateful firewalls.
 
 For the handler to work:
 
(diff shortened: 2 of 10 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -295,6 +296,13 @@ private function loadUserContext(ContainerBuilder $container, XmlFileLoader $loa
             $container->getDefinition($this->getAlias().'.user_context.logout_handler')
                 ->replaceArgument(1, $config['user_identifier_headers'])
                 ->replaceArgument(2, $config['match']['accept']);
+
+            $container->register($this->getAlias().'.user_context.logout.handler.session', ContextInvalidationLogoutHandler::class)
```
dbu: i'd prefer to declare this in the user context xml service definition file. https://symfony.com/doc/current/service_container/service_decoration.html also has an example for that.

src/Security/Http/Logout/ContextInvalidationLogoutSuccessHandler.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace FOS\HttpCacheBundle\Security\Http\Logout;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
+use Symfony\Component\Security\Http\Logout\SessionLogoutHandler;
+
+class ContextInvalidationSessionLogoutHandler implements LogoutHandlerInterface
```
dbu: should be final imho.

we should add a docblock with documentation to explain why we build this as decorator of the session logout handler rather than a stand-alone handler.

dbu:
@wesnick do you have time to continue on this? otherwise i will try to follow up, its one of the 2 things i want to have for the 2.2 release.

wesnick:
@dbu I don't have time to get to it this week, but probably next week.  Feel free to move it forward.  

A task still to take care of is the configuration scenario where you don't want to invalidate sessions in your logout handler but do want to invalidate user context cache (not sure the use case, but I suppose it could happen) needs some additional thought as it would not be possible to do that with the current direction of this PR.

NavyCoat:
@wesnick Will you have some time to look at this again? I'm looking for next version of this bundle.
Not sure if i can help with this issue. This is the only reason why i don't make any contribution to this.

wesnick:
I am looking at this now and will push some changes shortly.

src/Security/Http/Logout/ContextInvalidationSessionLogoutHandler.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace FOS\HttpCacheBundle\Security\Http\Logout;
+
+use FOS\HttpCacheBundle\UserContextInvalidator;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
+
+final class ContextInvalidationSessionLogoutHandler implements LogoutHandlerInterface
```
dbu: can we add a phpdoc that this replaces the session logout handler? i wonder if we should extend Symfony\Component\Security\Http\Logout\SessionLogoutHandler and call parent::logout. this increases the coupling, but i prefer that over not noticing when symfony changes that handler somehow...

wesnick:
> does out session handler alias win over the default service, regardless of whether security bundle or our bundle is loaded first?

I don't know the answer to this.  I don't know who I would ask either.  I could move to a compiler pass to be sure?

dbu:
i just looked at the code of container.php and found this:

```
// Attempt to retrieve the service by checking first aliases then
// available services. 
```

so its clear that its intentional that the alias wins over a service, so it won't accidentally change i think. can you please rebase to solve the merge conflict and extend the base logout handler and call its method, to make it explicit what we still do the same as that listener does?

wesnick:
I will update docs later this afternoon

wesnick:
tests fail on 7.2 with a segmentation fault, not sure what I can do about it

Resources/doc/reference/configuration/user-context.rst
```diff
@@ -139,26 +139,18 @@ or you will end up with mixed up caches.
 ~~~~~~~~~~~~~~~~~~
 
 The logout handler will invalidate any cached user hashes when the user logs
-out.
+out. This behavior applies to all stateful firewalls.
```
dbu: i would prefer a `.. note::` block with this plus the extra text you add below. something like: "The logout handler is active on all firewalls. If your application has multiple firewalls with different user contexts, you need to create your own custom invalidation handler. Be aware that Symfony's ``LogoutSuccessHandler`` places the ``SessionLogoutHandler`` before any configured logout handlers."

CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ Changelog
 
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
+  
+* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
+  ContextInvalidationSessionLogoutHandler.  The original handler was called 
+  after the invalidation of the session so did not correctly invalidate the
```
dbu: ... session so did not correctly invalidate the session ID in cache... => session, and thus did not invalidate the session it should have but a newly created one.

CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ Changelog
 
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
+  
+* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
+  ContextInvalidationSessionLogoutHandler.  The original handler was called 
+  after the invalidation of the session so did not correctly invalidate the 
+  session ID in cache.  You should remove the deprecated service 
+  `fos_http_cache.user_context.logout_handler` from the logout.handlers section
+  of your firewall.
```
dbu: ... firewall configuration.

wesnick:
@dbu I think this is complete.

CHANGELOG.md
```diff
@@ -27,10 +27,10 @@ Changelog
   
 * The ContextInvalidationLogoutHandler has been deprecated in favor of the 
   ContextInvalidationSessionLogoutHandler.  The original handler was called 
-  after the invalidation of the session so did not correctly invalidate the 
-  session ID in cache.  You should remove the deprecated service 
+  after the invalidation of the session, and thus did not invalidate the seesion
```
dbu: s/seesion/session/

CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ Changelog
 
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
+  
+* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
+  ContextInvalidationSessionLogoutHandler.  The original handler was called 
+  after the invalidation of the session, and thus did not invalidate the seesion
```
dbu: s/seesion/session/

CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ Changelog
 
 * User context is more reliable not cache when the hash mismatches. (E.g. after
   login/logout.)
+  
+* The ContextInvalidationLogoutHandler has been deprecated in favor of the 
+  ContextInvalidationSessionLogoutHandler.  The original handler was called
```
dbu: please use backticks around class names (this is markdown, single backtick is good).

double whitespace after the `.` on this line.

Resources/doc/reference/configuration/user-context.rst
```diff
@@ -147,18 +147,11 @@ For the handler to work:
...
-        firewalls:
-            secured_area:
-                logout:
-                    invalidate_session: true
-                    handlers:
-                        - fos_http_cache.user_context.logout_handler
+.. note::
+    The logout handler is active on all firewalls.  If your application has multiple firewalls
+    with different user context, you need to create your own custom invalidation handler. Be
+    aware that Symfony's ``LogoutSuccessHandler`` placea the ``SessionLogoutHandler`` before
```
dbu: s/placea/places/

src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
```diff
@@ -11,58 +11,33 @@
...
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
 
-class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
+/**
+ * @deprecated use ContextInvalidationSessionLogoutHandler in this same namespace as a replacement
+ *
+ * This handler is deprecated since in most cases the session has already been invalidated by the SessionLogoutHandler
```
dbu: lets be more affirmative: "This handler is deprecated because it never did what it was supposed to do. The session is already invalidated by the SessionLogoutHandler which is always the first logout handler executed."

wesnick:
@dbu I was out for a bit, now back and I updated PR per comments

dbu:
thanks a lot for fixing this problem, @wesnick !

PR #395, opened 2017-11-02
Support FQCN service names for Sensio Framework Extra Bundle

emodric:
From version 4.0 and up, `ControllerListener` service from `SensioFrameworkExtraBundle` has been renamed to use FQCN as service name, so this adds support for the old and new name.

`master` branch also needs this fix, however, the compiler pass has been renamed to `TagListenerPass`, so probably needs manual merge.

### DIFF
--- a/DependencyInjection/Compiler/TagSubscriberPass.php
+++ b/DependencyInjection/Compiler/TagSubscriberPass.php
@@ -25,12 +25,18 @@ class TagSubscriberPass implements CompilerPassInterface
     public function process(ContainerBuilder $container)
     {
         if (true === $container->getParameter('fos_http_cache.compiler_pass.tag_annotations')
-            && !$container->has('sensio_framework_extra.controller.listener')
+            && !$this->hasControllerListener($container)
         ) {
             throw new \RuntimeException(
                 'Tag support requires SensioFrameworkExtraBundle’s ControllerListener for the annotations. '
                 .'Please install sensio/framework-extra-bundle and add it to your AppKernel.'
             );
         }
+    }
+
+    private function hasControllerListener(ContainerBuilder $container)
+    {
+        return $container->has('sensio_framework_extra.controller.listener') ||
+            $container->has('Sensio\\Bundle\\FrameworkExtraBundle\\EventListener\\ControllerListener');
     }
 }


dbu:
hi, thanks for looking into this! the 1.3 branch is in maintenance mode and [does not allow symfony 4](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/composer.json#L26), i don't think we want to update this branch to ever support symfony 4. can you please redo this PR against the master branch instead?

emodric:
Hi @dbu!

It's about Symfony 4, but SensioFrameworkExtraBundle 4.x :)

I can redo the patch on the master branch, but then it would be great if 1.3 branch limited sensio extra bundle to 3.x, to disallow installing higher versions.

DependencyInjection/Compiler/TagSubscriberPass.php
```diff
@@ -25,12 +25,18 @@ class TagSubscriberPass implements CompilerPassInterface
...
                 'Tag support requires SensioFrameworkExtraBundle’s ControllerListener for the annotations. '
                 .'Please install sensio/framework-extra-bundle and add it to your AppKernel.'
             );
         }
     }
+
+    private function hasControllerListener()
+    {
+        return $container->has('sensio_framework_extra.controller.listener') ||
+            $container->has('Sensio\Bundle\FrameworkExtraBundle\EventListener\ControllerListener');
```
dbu: please use `\\` to make this strictly correct. 

in master we can use the ::class constant, but 1.3 unfortunately still supports php 5.3

DependencyInjection/Compiler/TagSubscriberPass.php
```diff
@@ -25,12 +25,18 @@ class TagSubscriberPass implements CompilerPassInterface
...
             throw new \RuntimeException(
                 'Tag support requires SensioFrameworkExtraBundle’s ControllerListener for the annotations. '
                 .'Please install sensio/framework-extra-bundle and add it to your AppKernel.'
             );
         }
     }
+
+    private function hasControllerListener()
+    {
+        return $container->has('sensio_framework_extra.controller.listener') ||
```
dbu: you need to pass the $container to this helper method.

emodric:
@dbu Rebased, fixed the `$container` param and FQCN.

Want me to do a PR on `master` branch or are you merging in manually?

emodric:
Thanks @dbu !

dbu:
tagged https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/1.3.11 for the 1.3 version. merging to master worked without conflicts, git understood that the file is the same and just renamed ;-)

The-Don-Himself:
Thanks @emodric , @dbu would we also be seeing these changes in versions 2.x? I had also commented possibly running into this https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/512

dbu:
@The-Don-Himself yes, i merged 1.3 to master to get this in 2.x as well. its here: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/Compiler/TagListenerPass.php

it will be in 2.2, which will be released when symfony 3.4 is out.

PR #405, opened 2017-11-18
Symfony 4 compatibility

adorjand:
Symfony 4 compatibility based on [this](https://symfony.com/blog/help…ing-prepare-for-symfony-4-bundle-support) article.

#SymfonyConHackday2017

* [ ] make test app kernel loadable with psr-3, add KERNEL_CLASS to phpunit (but keep KERNEL_DIR for testing older versions of symfony)
* [ ] make cache_manager public
* [ ] find solution for fos_http_cache.event_listener.tag, symfony_response_tagger, proxy_client.varnish,  - either get access without having to get it from the container, or make it public for the test env only
* [ ] `ServiceTest::testCanBeLoaded` => guess we should use the container builder instead of the compiled container in that test. most of those services need not be public probably.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,11 +73,7 @@ before_script:
   - composer update $COMPOSER_FLAGS
 
 script:
-  - if [ "$SYMFONY_DEPRECATIONS_HELPER_VALUE" != "" ]
-    then
-      SYMFONY_DEPRECATIONS_HELPER=$SYMFONY_DEPRECATIONS_HELPER_VALUE vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}
-    else
-      SYMFONY_DEPRECATIONS_HELPER=weak vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}
+  - if [[ "$SYMFONY_DEPRECATIONS_HELPER_VALUE" != "" ]]; then SYMFONY_DEPRECATIONS_HELPER=$SYMFONY_DEPRECATIONS_HELPER_VALUE vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}; else SYMFONY_DEPRECATIONS_HELPER=weak vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}; fi;
   - if [[ "$DOCCHECK" = true ]]; then make -C Resources/doc SPHINXOPTS='-nW' html; fi
   - if [[ "$DOCCHECK" = true ]]; then make -C Resources/doc spelling; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,10 +51,6 @@ matrix:
       env:
         - SYMFONY_VERSION='3.2.*'
         - SYMFONY_PHPUNIT_VERSION=5.5
-    - php: 7.1
-      env:
-        - DEPENDENCIES=beta
-        - SYMFONY_PHPUNIT_VERSION=5.5
     - php: hhvm
       dist: trusty
 
@@ -62,7 +58,6 @@ install:
   - if [[ "$DOCCHECK" = true ]]; then pip install -qr Resources/doc/requirements.txt --user; fi
 
 before_install:
-  - if [ "$DEPENDENCIES" = "beta" ]; then composer config minimum-stability beta; fi;
   - sh -c 'if [ "$SYMFONY_VERSION" != "" ]; then composer require --dev --no-update symfony/symfony=$SYMFONY_VERSION; fi;'
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ matrix:
       env:
         - SYMFONY_VERSION='3.4.*'
         - SYMFONY_PHPUNIT_VERSION=5.5
-        - $SYMFONY_DEPRECATIONS_HELPER_VALUE=52
+        - SYMFONY_DEPRECATIONS_HELPER_VALUE=52
     - php: 7.2
       env:
         - SYMFONY_VERSION='4.0.*'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -51,13 +51,18 @@ matrix:
       env:
         - SYMFONY_VERSION='3.2.*'
         - SYMFONY_PHPUNIT_VERSION=5.5
+    - php: 7.1
+      env:
+        - DEPENDENCIES=beta
+        - SYMFONY_PHPUNIT_VERSION=5.5
```
ddeboer: Symfony 4 is now released, so @adorjand can you please remove this Travis job?

.travis.yml
```diff
@@ -51,13 +51,18 @@ matrix:
...
     - php: hhvm
       dist: trusty
 
 install:
   - if [[ "$DOCCHECK" = true ]]; then pip install -qr Resources/doc/requirements.txt --user; fi
 
 before_install:
+  - if [ "$DEPENDENCIES" = "beta" ]; then composer config minimum-stability beta; fi;
```
ddeboer: Can be removed as well.

adorjand:
Sure, here you are. Thanks @ddeboer

dbu:
thanks for updating. however, now we have no build with symfony 3.4 nor symfony 4.0 anymore. may i ask you to add symfony 3.4 and symfony 4.0 to .travis.yml? otherwise we don't see if things actually work.

dbu:
oh, and don't worry about the styleci error. i just fixed that in master. you can rebase on master if you want, but for now its obvious that this branch does not change anything about the code style so we can also leave as is.

dbu:
we did not yet release a version of FOSHttpCache that allows symfony 4, so the symfony 4 build fails. 

and the 3.4 build fails as well. i guess step one would be to make the symfony deprecations helper output deprecation warnings. i suspect thats what happens in the 3.4 build, some warnings throwing exceptions in test and then assertions not showing the response they got but just saying that the response was not as expected... can you try to increase the output on one of the failing tests? i guess they are all the same pattern, so hopefully once we find the reason for one, the others are either solved as well, or at least we know how to solve.

dbu:
the "problem" is that we have two almost finished pull requests there. i'd love to get those in, it feels its like only cleanup left. lets set the deadline that next monday we release with whatever is in until then? https://github.com/FriendsOfSymfony/FOSHttpCache/milestone/5

dbu:
we have to fix 3.4 first anyways. ah, although some of that might be caused by not having the most recent version of the library which does some adjustments for symfony 3.4 deprecations. maybe change the requirement on http-cache to 2.1.x-dev until i managed to tag? then composer will automatically install the dev version of that, but leave other dependencies stable?

adorjand:
I hope I did it right

adorjand:
I didn't

composer.json
```diff
@@ -23,8 +23,8 @@
     "require": {
         "php": "^5.6.0||^7.0.0",
         "friendsofsymfony/http-cache": "^2.0.0",
```
dbu: here you need `^2.1.x-dev` for the time being

.travis.yml
```diff
@@ -64,7 +73,7 @@ before_script:
   - composer update $COMPOSER_FLAGS
 
 script:
-  - SYMFONY_DEPRECATIONS_HELPER=weak vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}
+  - if [[ "$SYMFONY_DEPRECATIONS_HELPER_VALUE" != "" ]]; then SYMFONY_DEPRECATIONS_HELPER=$SYMFONY_DEPRECATIONS_HELPER_VALUE vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}; else SYMFONY_DEPRECATIONS_HELPER=weak vendor/bin/simple-phpunit ${PHPUNIT_FLAGS}; fi;
```
dbu: we should just specify nothing for the helper, then we get to see the deprecations and can work on eliminating them.

adorjand: The deprecations can be seen after the last commit, mostly the private services are the problem

composer.json
adorjand: ok, I will change that too

.travis.yml
dbu: ah, right. great!
i'll add a todo list for the deprecations with some comments. any chance that you can work on those? (not sure how much you are willing to do, after all this just started as a small change during the symfony con hackday ;-) )

adorjand: In 'real world' I'm a web designer and front-end developer, but this should be no problem. I have a minimal to medium knowledge about back-end so it takes more time for me (reading docs, find correct solution, etc.), but if you want I can help in my free time. Anyway I think mostly will be only code refactoring.

dbu: great oportunity to learn something! and indeed this should be a good intro as its not about building something new but about small refactorings. just pick the ones you are comfortable with to start, and lets get the number of warnings down. hopefully the errors also go down while we do that. i might also do some pull requests against your branch to fix some of the things.

composer.json
```diff
@@ -22,9 +22,9 @@
     ],
     "require": {
         "php": "^5.6.0||^7.0.0",
-        "friendsofsymfony/http-cache": "^2.0.0",
-        "symfony/framework-bundle": "^2.8||^3.0",
-        "symfony/http-foundation": "~2.8.13||^3.1.6"
+        "friendsofsymfony/http-cache": "^2.1.x-dev",
```
dbu: looking at travis-ci, i think this should be `^2.1@dev` to allow composer to install a dev version. we now tell it to install at least the specific dev version, but not allow to install dev versions (afaik composer does not interpret the .x-dev to understand that it should install a dev version.

dbu:
i continue in #420 . @adorjand thanks for the start! if you want to help further, comment in #420 or do pull request against the symfony-4-compatibility branch in this repository.

PR #417, opened 2017-12-21
allow cache tag invalidation with symfony HttpCache

### LINKED ISSUE #403: adjust to symfony taggable cache support
see https://github.com/FriendsOfSymfony/FOSHttpCache/pull/376

notable around here we need to allow symfony client as well: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/FOSHttpCacheExtension.php#L403-L410

dbu:
fix #403

@Toflar is this all that is needed or am i missing something?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,15 @@ Changelog
 
 ### Added
 
+* You can now use cache tags and invalidate them with the Symfony `HttpCache`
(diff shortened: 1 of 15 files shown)


dbu:
ah, need to be able to configure the symfony proxy client too. adding that now...

Toflar:
Thank you @dbu!

PR #420, opened 2017-12-28
Symfony 4 compatibility

dbu:
Symfony 4 compatibility based on [this](https://symfony.com/blog/help…ing-prepare-for-symfony-4-bundle-support) article.

* [x] make test app kernel loadable with psr-3, add KERNEL_CLASS to phpunit (but keep KERNEL_DIR for testing older versions of symfony)
* [x] make cache_manager public
* [x] Update matthiasnoback/symfony-dependency-injection-test to 2.3 (that supports sf 4.0)
* [x] find solution for fos_http_cache.event_listener.tag, symfony_response_tagger, proxy_client.varnish,  - either get access without having to get it from the container, or make it public for the test env only
* [x] `ServiceTest::testCanBeLoaded` => guess we should use the container builder instead of the compiled container in that test. most of those services need not be public probably. 
* [x] Make some more services public in testing so that we can test them


composer.json
```diff
@@ -33,10 +33,10 @@
         "mockery/mockery": "0.9.*",
         "monolog/monolog": "*",
         "sensio/framework-extra-bundle": "^3.0",
-        "symfony/symfony": "^2.8||^3.0",
-        "symfony/phpunit-bridge": "^3.2",
-        "symfony/expression-language": "^2.4||^3.0",
-        "symfony/monolog-bundle": "^2.3||^3.0",
+        "symfony/symfony": "^2.8||^3.0||^4.0",
```
Tobion: It should not require symfony/symfony. That defeats the whole purpose of requiring the individual components. It's just require-dev but still wrong.

composer.json
```diff
@@ -23,8 +23,8 @@
     "require": {
         "php": "^5.6.0||^7.0.0",
         "friendsofsymfony/http-cache": "^2.1.0",
-        "symfony/framework-bundle": "^2.8||^3.0",
-        "symfony/http-foundation": "~2.8.13||^3.1.6"
+        "symfony/framework-bundle": "^2.8||^3.0||^4.0",
```
Tobion: IMO you can drop support for unmaintained versions like 3.0, 3.1 and 3.2

composer.json
dbu: indeed. we should use something like dunglas/symfony-lock in the ci instead, so that we see if our require deps are all that is needed.

tests/Functional/DependencyInjection/ServiceTest.php
```diff
@@ -45,3 +63,17 @@ public function testCanBeLoaded()
...
+class ServicesPublicPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        foreach ($container->getServiceIds() as $id) {
+            if (strncmp('fos_http_cache.', $id, 15)) {
+                continue;
+            }
+
+            $container->setDefinition($id, $container->getDefinition($id)->setPublic(true));
```
Tobion: Why does it set the same definition with the same id again? Just `$container->getDefinition($id)->setPublic(true)` should be enough.

src/Command/InvalidatePathCommand.php
```diff
@@ -39,7 +36,7 @@ class InvalidatePathCommand extends BaseInvalidateCommand
      */
     public function __construct(CacheManager $cacheManager = null, $commandName = 'fos:httpcache:invalidate:path')
     {
-        $this->commandName = $commandName;
+        static::$defaultName = $commandName;
```
Tobion: IMO it does not make sense to use the static property for this. The static property is only used before the command is instantiated. Setting it in the constructor doesn't do anything.

Tobion: I would also deprecate passing the commandName argument in the constructor.

dbu: yes, setting the name in the constructor was a weird idea of flexibility. i wanted to deprecate that, willdo.

about the static property: i understood that this is the sf 4 way to do the command name, with lazy services. i explicitly define the property so that this works with older symfony versions that do not have the property.

dbu:
the remaining test failures are all about https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/9a4875062f9dc30fd22c3950c8602325761c5844/src/EventListener/TagListener.php#L111 - debugging showed me that it only sees a request that is not a master request. which is odd, as the request is created as a master request... do you have any idea @Tobion? can we maybe look into this at the office today?

tests/Functional/EventListener/TagListenerTest.php
```diff
@@ -38,15 +38,24 @@ public function testAnnotationTagsAreInvalidated()
...
+        $mock->expects($this->at(0))
+            ->method('invalidateTags')
+            ->with(['all-items'])
+        ;
+        $mock->expects($this->at(1))
+            ->method('invalidateTags')
+            ->with(['item-123'])
+        ;
+        $mock->expects($this->once())
+            ->method('flush')
```
Tobion: > Expectation failed for method name is equal to <string:flush> when invoked 1 time(s).

That's the reported failure. And I don't understand what that expectation is meant to do. https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3dfbad1bdfc78d259a73359cc6131c4b7c31726a/src/EventListener/TagListener.php does not call `flush` on the CacheManager at all. So this cannot work.
Also this to me this `testTagsAreSetWhenCacheable` looks much more like a unit test and not a functional test. I don't see the point of using the container here and then setting a mock on it.
I thougt replacing services with mocks at runtime does not work with sf 4 anyway.

Tobion: Also I'm not sure if you can combine `$this->once()` and `$this->at(...)` in a single mock at all.

dbu: > I thougt replacing services with mocks at runtime does not work with sf 4 anyway.

It was re-enabled in https://github.com/symfony/symfony/pull/24418 - as long as they are public and not initialized.

once and at look bad together, but i think we have it working elsewhere. will check with `at()`, maybe thats it.

the idea of the test is that it would ensure the tags actually end up in the response - only unit tests are not telling me that. when i locally dump the type of request, i get told that the listener sees a sub request and thats why it does not flush.

src/Command/InvalidatePathCommand.php
dbu: setting the default on the static property now, and updated the doc to say using this parameter is deprecated

src/Resources/config/cache_manager_commands.xml
```diff
@@ -0,0 +1,25 @@
...
+        </service>
+
+        <service id="fos_http_cache.command.invalidate_regex" class="FOS\HttpCacheBundle\Command\InvalidateRegexCommand">
+            <argument type="service" id="fos_http_cache.cache_manager" />
+            <tag name="console.command" command="fos:httpcache:invalidate:regex"/>
+        </service>
+
+        <service id="fos_http_cache.command.refresh_path" class="FOS\HttpCacheBundle\Command\RefreshPathCommand">
+            <argument type="service" id="fos_http_cache.cache_manager" />
+            <tag name="console.command"/>
```
Tobion: this is missing the "command" attribute compared to the other command. but since you set the static property now, you don't need to specify the command name in the services anymore

XWB:
@dbu Any clue when `FOSHttpCacheBundle` 2.2 will be released?

dbu:
@XWB i am trying to wrap up some other things like https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/430 but hope that after that i can release.

one big issue is https://github.com/symfony/symfony/issues/25736#issuecomment-358791535 - if you have time to look into this and maybe come up with a solution, that would be great.

XWB:
@dbu Yes we suffered the same issue after upgrading to Symfony 3.4, but we hacked around it in the Varnish config:

```
sub vcl_backend_response {
    if (beresp.http.Cache-Control ~ "s-maxage" && beresp.http.Cache-Control !~ "s-maxage=0")
        set beresp.http.Cache-Control = regsub(beresp.http.Cache-Control, "private", "public");
        set beresp.http.Cache-Control = regsub(beresp.http.Cache-Control, "must\-revalidate, ", "");
        set beresp.http.Cache-Control = regsub(beresp.http.Cache-Control, "max-age=0, ", "");
    }
```

So if maxage is set, we set the page to public before Varnish caches it.

PR #421, opened 2018-01-11
Fix: AnonymousRequestMatcher doesn't check session name prefix

blankse:
I use eZ Platform and a anonymous with cookies (e. g. from piwik) but without the session cookie doesn't use the http cache.

The UserContextSubscriber in FOSHttpCache checks the session name prefix:
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/1.4/src/SymfonyCache/UserContextSubscriber.php#L110

The UserContextSubscriber with AnonymousRequestMatcher of the FOSHttpCacheBundle checks only the cookie count:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/EventListener/UserContextSubscriber.php#L108

So the sanity check failed:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/EventListener/UserContextSubscriber.php#L168


andrerom:
Unsure how it can be aligned, but we have a bit of a overlap with UserContextListener(detecting if anonymous and the setting): https://github.com/FriendsOfSymfony/FOSHttpCache/blob/e04f13c0190473bbea366a9038da91d596c34a1e/src/SymfonyCache/UserContextListener.php#L68

However UserContextListener is part of AppCache so can't read settings from config, only from what you init on custom AppCache yourself. (In eZ we set that here: https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/AppCache.php#L47)

blankse:
> But maybe we could provide service from AppCache to kernel/container for app logic to determine if user is anonymous to avoid the duplication?
Yes this would be nice. But I don't know a solution.

user_hash_header is also set twice until now:
FOSHttpCacheBundle/FOSHttpCache:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/DependencyInjection/Configuration.php#L653
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/1.4/src/SymfonyCache/UserContextSubscriber.php#L65
Overridden in ezpublish-http-cache:
https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/Resources/config/fos_http_cache.yml#L9
https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/AppCache.php#L50

dbu:
if there are other cookies, we have no way of knowing if those cookies are used by the backend to change the response or not. it could be a rememberme cookie, or the application could have stored any kind of information into that cookie and change the response based on the cookie.

therefore, the context lookup request has to be cleaned up to throw away all cookies that are considered unimportant for the backend (piwik/google analytics/etc). we cache the lookup request as "vary on cookie", thus if there are piwik/ga cookies that change with every request, it would always have to redo the context hash.

for symfony httpcache, we say this:
http://foshttpcache.readthedocs.io/en/latest/symfony-cache-configuration.html#cleaning-the-cookie-header => https://github.com/FriendsOfSymfony/FOSHttpCache/blob/2e75cd7a7e753995b5dc6d61937a319dbb933b25/src/SymfonyCache/UserContextListener.php#L126

maybe the "detect anonymous" feature still has a bug here, however. it should probably be invoked after the cookie has been cleaned. (also note that its a very unimportant thing. if the system is correctly set up, you should get one hash lookup for anonymous per cache lifetime for the hash lookup. all other anonymous users should hit the cache of the hash.)

andrerom:
This is true, we do the same in eZ Platform now @blankse, as described in doc here and in Symfony doc. ref: https://github.com/ezsystems/ezplatform/blob/master/doc/varnish/vcl/varnish4_xkey.vcl#L54-L65

And before we also did it, but only for user hash lookup.
Isn't AnonymousRequestMatcher involved mainly in user hash lookup? Or is it involved for all requests?

blankse:
The AnonymousRequestMatcher is used in all requests.
The request cookies will cleaned only in the user hash lookup.

For Anonymous Requests (no PHPSESSID/eZSESSID cookie) no user hash lookup will made and a fixed userHash is set:
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/2e75cd7a7e753995b5dc6d61937a319dbb933b25/src/SymfonyCache/UserContextListener.php#L167

blankse:
> if there are other cookies, we have no way of knowing if those cookies are used by the backend to change the response or not.

If a other cookie is used by the backend the anonymous mechanism should be avoid.
I see in version 2 it can be set to null.

The problem here is with anonymous mechanism: The FOSHttpCache UserContextSubscriber.php set a fix user hash (because it is no session cookie) und the FOSHttpCacheBundle UserContextSubscriber.php generate a hash (because the request has cookies). So the sanity check failed (fixed hash is not generated hash). 

After this fix, the FOSHttpCacheBundle UserContextSubscriber.php don't generate a hash and the sanity check is valid (`!is_null($this->hash)`)

In the best case the same logic is used by both classes. But I have no idea how.

blankse:
I create a alternative PR: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/425
There the isAnonymous logic is shared from FOSHttpCache.

The session_name_prefix configuration still needs to be set twice.

dbu:
i prefer the refactoring approach you do in #425, will close this PR.

PR #422, opened 2018-01-12
Passed correct configuration for Nginx

hpatoio:


hpatoio:
Build fails for `HHVM` with this message

```
1) FOS\HttpCacheBundle\Tests\Unit\EventListener\UserContextListenerTest::testOnKernelRequest
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'max-age=0, no-cache, private'
+'max-age=1, no-cache, private'
/home/travis/build/FriendsOfSymfony/FOSHttpCacheBundle/tests/Unit/EventListener/UserContextListenerTest.php:69
```

I don't think is related to my changes.

dbu:
ouch. thanks a lot for this fix. i guess we also miss a test if we do not notice this in the test suite...

master is the upcoming 2.2 release. i created 2.1 from the 2.1.1 tag and cherry-picked this commit into the 2.1 branch: cbf34498261c47621a6e9f64a072b1fad7a3ca7a .  (changing the target in github did not have the desired effect). now i tagged 2.1.2 with the fix, and will merging 2.1 to master shortly

PR #425, opened 2018-01-12
Use AnonymousRequestMatcher from FOSHttpCache

blankse:
Requires https://github.com/FriendsOfSymfony/FOSHttpCache/pull/402
Alternative to https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/421


andrerom:
Looks ok with me, simplifies things a bit.
wdyt @dbu ?

blankse:
@dbu I think it't not a feature. It's a bug. It makes no sense to use different logics here.

The sanity check fails if one logic think the user is an anonymous and the other think he is not an anonymous:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/EventListener/UserContextSubscriber.php#L168

dbu:
yeah, its a bit on the edge. the UserContextSubscriber is only relevant for the symfony HttpCache, not for users on varnish cache. the refactoring has a risk of having side effects especially for varnish users. then again, the anonymous handling is not active by default with varnish. and i asked you to refactor rather than do the smallest possible change. (talking to myself here :-D ) anyways, i already merged the thing in the library to 1.4, might as well continue here. 

however, then we need to solve the issue that the user_identifier_header argument is currently lost. please keep that argument (i guess that means adding it to the library - as we did not release there yet, i would see an $options parameter that accepts the 2 options session prefix and user identifier header. please use the OptionsResolver to provide defaults.

UserContext/AnonymousRequestMatcher.php
```diff
@@ -1,47 +0,0 @@
...
-
-namespace FOS\HttpCacheBundle\UserContext;
-
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\RequestMatcherInterface;
-
-/**
- * Matches anonymous requests using a list of identification headers.
- */
-class AnonymousRequestMatcher implements RequestMatcherInterface
```
dbu: for BC, we need to keep this class around. i think the best is to make it extend the one from the library but keep the constructor signature, and forward the headers as the options array. and trigger a deprecation warning.

imo its ok to make the service definition use the library matcher directly. my thing is for people who use this class to define their own service.

dbu:
another thing about BC and making this work: the user identifier headers currently include "Cookie" - i think the default configuration for session prefix needs to be `false` and when its set to a string, a configuration cleanup step should remove the "Cookie" header from the list of user identifier headers. otherwise as soon as there is any cookie in the request, the detection fails.

please add the extra basic auth headers from the library to the symfony configuration defaults. those are inconsistent as well atm, we only look for Authorization but not HTTP_AUTHORIZATION or PHP_AUTH_USER.

dbu:
sorry to be a nitpick but BC is really important for users of this library, so i want to avoid regressions as much as possible.

blankse:
I'm sorry for the long delay.

I changed the PR and create a new PR in the HTTPCache library: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/408

> another thing about BC and making this work: the user identifier headers currently include "Cookie" - i think the default configuration for session prefix needs to be false and when its set to a string, a configuration cleanup step should remove the "Cookie" header from the list of user identifier headers. otherwise as soon as there is any cookie in the request, the detection fails.

I do not agree with that. We should use the same configuration in HTTPCache and HTTPCacheBundle by default.

What do you think about my solution?

UserContext/AnonymousRequestMatcher.php
```diff
@@ -11,27 +11,27 @@
...
+class AnonymousRequestMatcher extends BaseAnonymousRequestMatcher
 {
-    private $userIdentifierHeaders;
-
-    /**
-     * @param array $userIdentifierHeaders List of request headers that authenticate a non-anonymous request
-     */
-    public function __construct(array $userIdentifierHeaders)
+    public function __construct(array $options = array())
```
dbu: we need some BC code here that detects if the options actually are the new options or a list of headers. (check for the keys that options can have and if such a key is present, its options. otherwise you need to create the options array for the parent, with the user identifier headers as the value that is set

UserContext/AnonymousRequestMatcher.php
```diff
@@ -11,27 +11,27 @@
...
 
+use FOS\HttpCache\UserContext\AnonymousRequestMatcher as BaseAnonymousRequestMatcher;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\RequestMatcherInterface;
 
 /**
  * Matches anonymous requests using a list of identification headers.
  */
-class AnonymousRequestMatcher implements RequestMatcherInterface
+class AnonymousRequestMatcher extends BaseAnonymousRequestMatcher
```
dbu: please add a `@deprecated` annotation saying to use the class from the library instead

UserContext/AnonymousRequestMatcher.php
```diff
@@ -11,27 +11,27 @@
...
-        $this->userIdentifierHeaders = $userIdentifierHeaders;
+        @trigger_error(
+            'AnonymousRequestMatcher of HttpCacheBundle is deprecated. '.
+            'Use AnonymousRequestMatcher of HttpCache library.',
+            E_USER_DEPRECATED
+        );
+        parent::__construct($options);
     }
 
     public function matches(Request $request)
```
dbu: it should be possible to remove this method completely, or not?

dbu:
@blankse any chance that you can wrap this one up?

blankse:
@dbu I changed the PR

dbu:
we should add something to the changelog. i just noticed that if you set the `Cookie` header in the list of identifying headers, we keep the old behaviour of any cookie being set making the request not anonymous. we could mention that in the documentation.

btw, this is half a note to myself. if you want to work on it, great, but otherwise i can take it from here and wrap things up.

blankse:
@dbu Would be good if you could do that ;)

blankse:
@dbu I fixed the tests.

I noticed that the logout handler also use the user identifier headers.
I think the current logic requires the Cookie value. WDYT?

dbu:
see #434

PR #426, opened 2018-01-15
Improved commands

hpatoio:
Improved commands


dbu:
i guess this just happened to work because symfony autodetected the command in the Command folder, and it has no required constructor arguments... thanks for the fix, will avoid some wtf.

are you using the command name parameter? i recently discussed with somebody that we should probably drop that parameter and just hardcode the name. its not like the commands are reusable bits of code.

PR #427, opened 2018-01-19
[WIP][RFC] Fastly support

hpatoio:
Adding Fastly support. At the moment only tag invalidation. Goes with https://github.com/FriendsOfSymfony/FOSHttpCache/pull/403


dbu:
will need to be updated once https://github.com/FriendsOfSymfony/FOSHttpCache/pull/451 is wrapped up

composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^5.6.0||^7.0.0",
-        "friendsofsymfony/http-cache": "^2.1.0",
+        "friendsofsymfony/http-cache": "dev-fastly_support",
```
andrerom, line 25:
```suggestion
        "friendsofsymfony/http-cache": "^2.9.0",
```

andrerom (commented):
+1

Besides the composer change suggested, and a rebase, anything else needed here @dbu?

src/DependencyInjection/Configuration.php
```diff
@@ -372,6 +372,32 @@ private function addProxyClientSection(ArrayNodeDefinition $rootNode)
...
+                        ->arrayNode('fastly')
+                            ->children()
+                                ->scalarNode('authentication_token')
+                                    ->info('Fastly auth token.')
+                                ->end()
+                                ->scalarNode('service_identifier')
+                                    ->info('Fastly service identifier.')
+                                ->end()
+                                ->scalarNode('soft_purge')
+                                    ->info('Set soft purge mode.')
```
andrerom, line 384:
```suggestion
                                ->booleanNode('soft_purge')
                                    ->defaultTrue()
                                    ->info('Disable soft purge mode (enabled by default)')
```

src/Resources/config/fastly_proxy.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="fos_http_cache.test.proxy_server.fastly"
+                 class="FOS\HttpCache\Test\Proxy\FastlyProxy">
```
dbu, line 9: i don't think a FastlyProxy class exists.

we use this pattern to start a varnish / nginx instance during testing. i don't think we can do that for fastly. 

can we remove this config file and the DI method that loads it?

dbu (commented):
uh, i completely forgot to look at this again after fastly support was merged into FOSHttpCache

this looks good, i think there are some things to be removed.

and can you please add a test to ConfigurationTest? testSupportsFastly, similar to testSupportsVarnish and the others.

andrerom:
@hpatoio You ok with doing the fix-ups or would you like us to take over on it?

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -318,6 +318,9 @@ private function loadProxyClient(ContainerBuilder $container, XmlFileLoader $loa
         if (isset($config['nginx'])) {
             $this->loadNginx($container, $loader, $config['nginx']);
         }
+        if (isset($config['fastly'])) {
+            $this->LoadFastly($container, $loader, $config['fastly']);
```
proeinfo, line 322: typo

proeinfo (commented):
To allow tag support, fastly also needs to be included here: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/4d8e63228ef99270e266e4032a2965cb8043afd8/src/DependencyInjection/FOSHttpCacheExtension.php#L454

and with changes in Master, included here to disregard `http.servers`: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/Configuration.php#L492

src/DependencyInjection/FOSHttpCacheExtension.php
andrerom: Hi @proeinfo

So did you try this as well? Would be great if that is the case.
Either way, would you be able to use the github suggest feature to suggest fixes to the two things you point out here?

proeinfo: Hi @andrerom

I'm working on getting this to work, will post patch once that happens.

proeinfo:
An additional issue: since Fastly doesn't use proxyservers for invalidating, the HttpDispatcher's approach of running through proxy servers (done in the `fanOut()` method) won't work.

A simple solution is to extend the HttpDispatcher and remove any mentions of servers and baseUris.

Attached example of such a class.
[FastlyDispatcher.txt](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/files/5457780/FastlyDispatcher.txt)

This would just require that the Fastly proxy client is instantiated with the FastlyDispatcher instead:
```php
$some_example_tags = [
  'product1',
  'product2',
];
$httpDispatcher = new \FOS\HttpCache\ProxyClient\FastlyDispatcher();
$options = [
    'service_identifier' => '...',
    'authentication_token' => '...',
    'soft_purge' => true,
];
$fastly = new \FOS\HttpCache\ProxyClient\Fastly($httpDispatcher, $options);
$cacheInvalidator = new \FOS\HttpCache\CacheInvalidator($fastly);
$cacheInvalidator->invalidateTags($some_example_tags)->flush();
```

proeinfo:
Additionally, since Fastly uses space separated surrogate keys, the default comma-separated tag header formatter won't work.
Again it may make sense to create a new class for Fastly.

Example implementation:
```php
<?php

namespace FOS\HttpCache\TagHeaderFormatter;

class FastlyTagHeaderFormatter implements TagHeaderFormatter
{
    const FASTLY_TAG_HEADER = 'Surrogate-Key';
    const FASTLY_TAG_HEADER_GLUE = ' ';

    /**
     * @var string
     */
    private $headerName = self::FASTLY_TAG_HEADER;

    /**
     * @var string
     */
    private $glue = self::FASTLY_TAG_HEADER_GLUE;

    /**
     * {@inheritdoc}
     */
    public function getTagsHeaderName()
    {
        return $this->headerName;
    }

    /**
     * {@inheritdoc}
     */
    public function getTagsHeaderValue(array $tags)
    {
        return implode($this->glue, $tags);
    }
}

```

From there its a simple matter of overriding the service id in the user Symfony application:
```yaml
services:
    fos_http_cache.tag_handler.header_formatter:
        class: FOS\HttpCache\TagHeaderFormatter\FastlyTagHeaderFormatter

```

PR #434, opened 2018-03-23
Use AnonymousRequestMatcher from FOSHttpCache

dbu:
continue #425


andrerom:
Is there some BC notes that should be mentioned in regards to user_identifier_headers change? or is logic the same but using session_name_prefix differently?

dbu:
i am cleaning things up to not have BC breaks in this, but just make it possible to use the new features.

andrerom:
ok

Issue #433, opened 2018-03-20
work around symfony 4 auto-adding cache control headers


dbu:
Symfony 4 sets cache control headers when there is a session. See https://github.com/symfony/symfony/issues/25736#issuecomment-358791535 for details.

This will be a problem for the user context caching, because there we explicitly want to cache content from logged in users.

We need to find a solution that respects intended cache control instructions but does not get destroyed by the symfony automagic.

Contao did a workaround like this: https://github.com/contao/core-bundle/pull/1388

dbu:
fixed in #435 . thanks a lot @Toflar !

PR #435, opened 2018-03-23
Decorate default Symfony session listener for Symfony 3.4+

Toflar:
Fix for https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/433.

### DIFF
--- a/src/EventListener/SessionListener.php
+++ b/src/EventListener/SessionListener.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionListener;
+
+/**
+ * Decorates the default Symfony session listener.
+ *
+ * The default Symfony session listener automatically makes responses private
+ * in case the session was started. This kills the user context feature of
+ * FOSCache. We disable the default behaviour only if the user context header
+ * is part of the Vary headers to reduce the possible impacts on other parts
+ * of your application.
+ *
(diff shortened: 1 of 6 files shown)


src/EventListener/SessionListener.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionListener;
+
+/**
+ * Decorates the default Symfony session listener.
+ *
+ * The default Symfony session listener automatically makes responses private
+ * in case the session was started. This kills the user context feature of
+ * FOSCache. We disable the default behaviour only if the user context header
```
dbu: s/FOSCache/FOSHttpCache

src/EventListener/SessionListener.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @var array
+     */
+    private $userIdentifierHeaders;
+
+    public function __construct(BaseSessionListener $inner, string $userHashHeader, array $userIdentifierHeaders)
+    {
+        $this->inner = $inner;
+        $this->userHashHeader = strtolower($userHashHeader);
+        $this->userIdentifierHeaders = array_map('strtolower', $userIdentifierHeaders);
```
dbu: given that this is executed on every single request, i think we should do this in the DI extension class instead.

Toflar:
Comments addressed, tests fixed :)

dbu:
@andrerom could you maybe have a look on this? and do you have a setup where you can test if it actually works in a real environment?

dbu:
we manually verified that the listener does what we expect from it.

we will proceed, worst case there will be a 2.2.1 bugfix release...

emodric:
@dbu How will this handle Symfony 4.1 where the inner listener also does some other stuff in addition to setting caching headers: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/EventListener/AbstractSessionListener.php#L70-L97 ?

dbu:
argh. we went back and forth between calling session->save ourselves and then call the parent, or doing what we do now. we decided for this in the end because things would otherwise break if the SaveSessionListener would start doing additional things.

as they found it acceptable to call `save` in the session listener, we should thus do the same. and have to, as currently the session is not saved anymore with sf 4.1

dbu:
i created #437 and once somebody did that, will release 2.2.1

thanks a lot for having verified that @emodric !

dbu:
@emodric can you check #438 please?

PR #438, opened 2018-03-27
adjust to session listener changes in symfony 4.1

### LINKED ISSUE #437: Adjust session listener decorator for Symfony 4.1
We need to mimik more of the behaviour of the AbstractSessionListener as it was adjusted for 4.1 and call session->save ourselves and then call the inner in all cases. 

See https://github.com/symfony/symfony/pull/25836

dbu:
The SessionListener in Symfony 4.1 is changed and does the `$session->save()` itself.

Fix #437

### DIFF
--- a/src/EventListener/SessionListener.php
+++ b/src/EventListener/SessionListener.php
@@ -11,7 +11,9 @@
 
 namespace FOS\HttpCacheBundle\EventListener;
 
+use Psr\Container\ContainerInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionListener;
@@ -20,15 +22,22 @@ use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionLis
  * Decorates the default Symfony session listener.
  *
  * The default Symfony session listener automatically makes responses private
- * in case the session was started. This kills the user context feature of
+ * in case the session was started. This breaks the user context feature of
  * FOSHttpCache. We disable the default behaviour only if the user context header
- * is part of the Vary headers to reduce the possible impacts on other parts
+ * is part of the Vary headers to reduce the risk of impacts on other parts
  * of your application.
+ *
+ * @internal
  *
  * @author Yanick Witschi <yanick.witschi@terminal42.ch>
  */
 final class SessionListener implements EventSubscriberInterface
 {
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
     /**
      * @var BaseSessionListener
      */
@@ -45,12 +54,18 @@ final class SessionListener implements EventSubscriberInterface
     private $userIdentifierHeaders;
 
     /**
+     * @param ContainerInterface  $container             To allow to be extra lazy
      * @param BaseSessionListener $inner
      * @param string              $userHashHeader        Must be lower-cased
      * @param array               $userIdentifierHeaders Must be lower-cased
      */
-    public function __construct(BaseSessionListener $inner, string $userHashHeader, array $userIdentifierHeaders)
-    {
+    public function __construct(
+        ContainerInterface $container,
+        BaseSessionListener $inner,
+        string $userHashHeader,
+        array $userIdentifierHeaders
+    ) {
+        $this->container = $container;
         $this->inner = $inner;
         $this->userHashHeader = $userHashHeader;
         $this->userIdentifierHeaders = $userIdentifierHeaders;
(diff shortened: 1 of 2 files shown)


src/EventListener/SessionListener.php
```diff
@@ -45,12 +54,18 @@
     private $userIdentifierHeaders;
 
     /**
+     * @param ContainerInterface  $container             To allow to be extra lazy
      * @param BaseSessionListener $inner
      * @param string              $userHashHeader        Must be lower-cased
      * @param array               $userIdentifierHeaders Must be lower-cased
      */
-    public function __construct(BaseSessionListener $inner, string $userHashHeader, array $userIdentifierHeaders)
```
Toflar: You will have to adjust the extension as the argument indexes do not match anymore now.

src/EventListener/SessionListener.php
```diff
@@ -20,15 +22,22 @@
...
- * in case the session was started. This kills the user context feature of
+ * in case the session was started. This breaks the user context feature of
  * FOSHttpCache. We disable the default behaviour only if the user context header
- * is part of the Vary headers to reduce the possible impacts on other parts
+ * is part of the Vary headers to reduce the risk of impacts on other parts
  * of your application.
  *
+ * @internal
+ *
  * @author Yanick Witschi <yanick.witschi@terminal42.ch>
```
Toflar: Your name is missing :)

dbu:
@nicolas-grekas could you maybe have a look at this? for the user context caching, we explicitly want to have cache headers on responses that have a session. with the changes in symfony 4.1, this seems like the only way of achieving this.

src/EventListener/SessionListener.php
```diff
@@ -67,15 +82,30 @@ public function onKernelResponse(FilterResponseEvent $event)
             return;
         }
 
+        // copied from the Symfony SessionListener, to avoid breaking the lazyness
```
Toflar: Tests are missing on purpose I suppose?

src/EventListener/SessionListener.php
```diff
@@ -67,15 +82,30 @@ public function onKernelResponse(FilterResponseEvent $event)
...
 
-        // Call default behaviour if it's an irrelevant request for the user context
-        if (0 === count(array_intersect($varyHeaders, $relevantHeaders))) {
-            $this->inner->onKernelResponse($event);
+        if (0 < count(array_intersect($varyHeaders, $relevantHeaders))
+            && ($session->isStarted() || ($session instanceof Session && $session->hasBeenStarted()))
+        ) {
+            $session->save();
+
+            if ($session instanceof Session && $session->hasBeenStarted()) {
```
emodric: Shouldn't we return always if Vary header check is statisfied? It seems to me that with this, if the check is satisfied, but for anomymous users without session, the original listener still gets called, right?

emodric:
I guess we should check for Sf 4.1 and save the sesion only in that case?

Toflar:
> I guess we should check for Sf 4.1 and save the sesion only in that case?

Doesn't hurt. In 3.4 and 4.0 there's a `SessionSaveListener? That does nothing else than just saving the session which does nothing in case it was already saved :)

src/EventListener/SessionListener.php
dbu: yep, wanted some feedback on the general approach before i adjust DI and the tests.

src/EventListener/SessionListener.php
dbu: that is on purpose. i want to interfere as little as possible with the default behaviour. the only case where calling the parent after having done `$session->save()` would result in a problem is if the `hasBeenStarted` flag is present - that seems to try to work around the `save` and still detect if a session was open and has been saved...

Toflar:
As much as I agree with @nicolas-grekas that Symfony should set the response to `private` by default if the session is started, I think it's way too hard to disable this behaviour. Maybe a simple second boolean parameter in that listener that is set to `true` by default would be enough so we could just set that to `false` instead of decorating etc.?

emodric:
@Toflar Yep, agreed. This already looks like a wild goose chase. They change something, we fix it, and so on and so on.

Toflar:
Or even more flexible like this:

```php
if ($session->isStarted() || ($session instanceof Session && $session->hasBeenStarted())) {

    if ($response->headers->has('Symfony-NoPrivateIfSessionStarted')) {
        $response->headers->remove('Symfony-NoPrivateIfSessionStarted');
    } else {
        $response
            ->setPrivate()
            ->setMaxAge(0)
            ->headers->addCacheControlDirective('must-revalidate');
     }
}
```

Because there are multiple setups out there that wish to be able to disable this feature without having to do heavy lifting code. That one would easily solve all the issues.

Toflar:
PR is ready at https://github.com/symfony/symfony/pull/26681

nicolas-grekas:
But any new feature on Symfony would not be applicable to this bundle, unless you plan to make it support Symfony 4.1 only.

Toflar:
User context would be 4.1+ only. And my PR only disables the auto-private-thing. It leaves everything else as is.

dbu:
The solution we merged and tagged a few days ago works for symfony 3.4 and 4.0, but will break session saving with symfony 4.1 because the SaveSessionListener is no longer a separate step. We did not call the inner listener because it used to only overwrite the cache headers to private - now we need a different solution and `Session::hasBeenStarted` does not make it easier...

But lets discuss on https://github.com/symfony/symfony/pull/26681 first and then see what remains to be done here.

Toflar:
Yeah this listener is really causing trouble. A per response option to disable it would solve all our problems.

PR #439, opened 2018-04-05
adjust session listener to symfony 4.1

### LINKED ISSUE #437: Adjust session listener decorator for Symfony 4.1

dbu:
Use https://github.com/symfony/symfony/pull/26681 to make the session listener work with Symfony 4.1

Fix #437
Replaces #438

### DIFF
--- a/src/EventListener/SessionListener.php
+++ b/src/EventListener/SessionListener.php
@@ -19,11 +19,14 @@ use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionLis
 /**
  * Decorates the default Symfony session listener.
  *
- * The default Symfony session listener automatically makes responses private
- * in case the session was started. This kills the user context feature of
- * FOSHttpCache. We disable the default behaviour only if the user context header
- * is part of the Vary headers to reduce the possible impacts on other parts
- * of your application.
+ * Since Symfony 3.4, the default Symfony session listener automatically
+ * overwrites the Cache-Control headers to `private` in case the session has
+ * been started. This destroys the user context feature of FOSHttpCache.
+ * Since Symfony 4.1, there is a header we can set to skip this behaviour. We
+ * set that header in UserContextListener.
+ * For Symfony 3.4 and 4.0, we decorate the listener to only call the default
+ * behaviour if `Vary` contains neither the context hash header nor any of the
+ * user identifier headers, to avoid impacts on other parts of your application.
  *
  * @author Yanick Witschi <yanick.witschi@terminal42.ch>
  */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+2.2.1 (unreleased)
+------------------
+
+### Fixed
+
+* Adjust user context listener to handle Symfony 4.1 breaking behaviour change.
+
 2.2.0
 -----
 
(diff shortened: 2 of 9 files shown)


dbu:
@Toflar would you have a moment to review this one?

dbu:
@emodric can you also check if this makes sense?

emodric:
@dbu Looks good to me :+1:

Although, I was confused at first at why do you disable the original listener even on versions lower than Symfony 3.4 (`if (version_compare(Kernel::VERSION, '4.1', '<'))`), but then I remembered that the decorator is only active for Symfony >= 3.4 in a compiler pass. This could be made more explicit maybe: to do the version check in one place.

src/EventListener/SessionListener.php
```diff
@@ -73,9 +76,18 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+        }
+
+        if (version_compare(Kernel::VERSION, '4.1', '<')) {
+            // For Symfony 3.4 and 4.0, do not let the inner listener destroy caching. The session will be saved in a separate listener.
+            return;
         }
 
-        // noop, see class description
+        // For Symfony 4.1+, we need to call the inner listener to save the session. Set the header to disable overwriting caching headers.
+        $event->getResponse()->headers->set(AbstractSessionListener::NO_AUTO_CACHE_CONTROL_HEADER, 1);
```
Toflar: I don't think this should be set here. I think we should not decorate the listener at all in Symfony 4.1+ but instead set that header where it belongs: The `UserContextListener`. Imho it can always be added, even for lower Symfony versions because the header doesn't harm at all if it is being sent to external but if you really want to omit that you can add the version_compare there.

So basically imho we should only register/decorate this listener when Symfony 3.4 or Symfony 4.0.
And then I think - for the sake of your time and stress level - you should really think about dropping < 4.1 support in the future 😄

emodric: > So basically imho we should only register/decorate this listener when Symfony 3.4 or Symfony 4.0.

Yep, this goes inline with my suggestion to keep the version check at one place. It can be further extended to only decorate on Sf 3.4 and 4.0.

> And then I think - for the sake of your time and stress level - you should really think about dropping < 4.1 support in the future

I hope this won't happen before 3.0, though :)

dbu:
thanks for the input. indeed, thats much better. i now add the header in the UserContextListener and the SessionListener is purely a workaround for SF 3.4 and 4.0.

good like this, or any more input?

emodric:
Looks good! :+1:

Toflar:
Perfect, yes 👍 Maybe we should add a unit test covering that case in the `UserContextListener`? Or did you leave that out on purpose because 4.1 is not released yet? 😄

dbu:
got stuck trying to build that test... hope to have something soon.

@emodric i'd rather not maintain the 1.x branch anymore. the 1.3 branch does not support symfony 4. if you really need the 1.3 branch with symfony 3.4 and can't upgrade to FOSHttpCacheBundle 2, i can merge a pull request on the 1.3 branch for this, if you port it. might also safe other people who are stuck on version 1 and update symfony to 3.4.

emodric:
Thanks! I'll create a PR with the decorator, I would need it for Symfony 3.4, yes.

dbu:
tagged as 2.2.1

PR #440, opened 2018-04-13
Allow direct routing for Symfony proxy purge requests

Toflar:
Complementary PR for https://github.com/FriendsOfSymfony/FOSHttpCache/pull/419

TODO
* [x] Adjust to FOSHttpCache changes
* [x] Add documentation
* [x] Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+2.3.0
+-----
+
+### Added
+
+* You can now configure the Symfony proxy client to directly call the
+  kernel for invalidation requests instead of executing real requests
+  and thus reduce complexity for single node Symfony applications.
+  Use the new configuration `proxy_client.symfony.use_kernel_dispatcher`
+  and follow the setup instructions in both, the bundle and the library
+  to configure your kernel accordingly.
+
 2.2.0
 -----
 
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -149,6 +149,16 @@ symfony
 
 For ``servers``, ``base_url``, ``tags_header`` and ``header_length``, see above.
 
+
+.. versionadded:: 2.3
+
+    You can omit the whole ``http`` configuration and use ``use_kernel_dispatcher: true``
+    instead. This will call the kernel directly instead of executing a real HTTP request.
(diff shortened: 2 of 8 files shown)


dbu:
this will be a new feature, so we will have to bump the branch-alias to 2.3, and also require the FOSHttpCache library in a new minor version so that class exists.

Toflar:
Regarding the docs: They seem outdated, no? `class AppCache extends EventDispatchingHttpCache` for example. I think that is a trait now? Maybe that should be cleaned up first before we document the new feature (not sure if we need to document a lot here anyway because the proxy configuration is already properly documented in the library and all that's left here is to enable it using a configuration flag so I'd probably just link).

Redgarding the config: Do you know by chance how I can make the `http` config section in the bundle conditional? I mean it has to be there except for if you set `use_kernel_dispatcher` to `true`.

Toflar:
No idea why travis doesn't run here but ready for review :)

dbu:
looks good to me!

- please add documentation in the configuration reference: http://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/proxy-client.html#symfony
- can you please add some tests for the unhappy case when configuration is incomplete, to be sure we do get the error if server is not configured?

Toflar:
Done. Still not sure about the docs references but I hope you can fix them if they're wrong 😃

dbu:
pleaes change the branch alias in composer.json, and also change the minimum version of http-cache to 2.3 so that we are sure that we have the kernel dispatcher available.

Toflar:
Done :)

dbu:
i solved the merge conflict (in the changelog) and moved the KernelDispatcher documentation of FOSHttpCache into its own subsection in https://github.com/FriendsOfSymfony/FOSHttpCache/commit/cf09b88153b09e692aba9aa6ac98dbb50f0fc837 . i tweaked the documentation a bit to hopefully be more clear what needs to be done where, and merged it as 2c337377f7bf2b6bf6aef16aedee215f45f390fa

thanks for the contribution. want to write a blogpost about this? i think a fully fleshed out example would be useful.

dbu:
oh, and for the record about cross project links: we need a label to link to, its neither page name nor heading name: https://github.com/FriendsOfSymfony/FOSHttpCache/commit/59a658246d8565fb52b365664f13e623cafc254a

Toflar:
Thanks! I'm not blogging (yet).

PR #444, opened 2018-04-18
Fix session_listener decoration when session is not enabled

Tobion:
#435 breaks the app when session is not used

This requires the frameworkbundle to be registered before the httpcachebundle. But might be good enough.


dbu:
this change breaks a functional test. are you sure its correct like this? or do we need a compiler pass to only do it at the end?

Tobion:
Yeah it's never true. Needs a compiler pass.

PR #445, opened 2018-04-20
when sessions are not enabled, the listener does not exist either

dbu:
fix problem introduced in #435, replace #444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+2.2.2
+-----
+
+### Fixed
+
+* Fix session_listener decoration when session is not enabled.
+
 2.2.1
 -----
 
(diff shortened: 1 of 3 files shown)


dbu:
@Tobion can you please check if this fixes the problem you encountered?

Tobion:
Works 👍

emodric:
@dbu @Tobion Does this need to be backported to 1.4 branch too?

emodric:
1.3 cannot be installed on Symfony 4.x, so that's not an issue. That's why I didn't include #439 in the first place.

The only open question is about Symfony 3.4.

dbu:
ah good. but this change is relevant for 3.4 as well.

the listener expects the symfony session listener to exist. if sessions are not enabled, this is not the case, and container can't be compiled.

emodric:
Okay, do you want me to backport this as well?

dbu:
if you have the time, sure. i don't want to actively maintain version 1 anymore, but as the rest has been backported it would make sense to now also fix the regression (of not working without sessions anymore).

emodric:
Okay. Will do it in a day or two! :+1:

Issue #431, opened 2018-03-13
Add support for Varnish xkey


dbu:
Make the features of https://github.com/FriendsOfSymfony/FOSHttpCache/pull/332 available through configuration.

The xkey setting should be shared between ResponseTagger and Varnish client to change all headers to the correct default.

andrerom:
Could you expand on how it should be done, in detail? _(point to existing settings this can be modeled after, expand on what the setting should be named and how it should behave)_

dbu:
i didn't look into it in detail yet. looking at the Configuration class, i see there is the `addTagSection` method for the `tags` part of the configuration. i think this is where the config should go. something like `tag.mode` to be consistent with the naming in FOSHttpCache.
we can then add a new validate function at the top of Configuration to check that if the tags method is xkey, the client indeed is varnish, and change the appropriate defaults of the varnish configuration to take into account whether the xkey or ban method is used for tags, and set up the response tagger accordingly.
i added some of that default behaviour in FOSHttpCache already - maybe we can simply change some defaults to be null instead of repeat the default value from the library to get FOSHttpCache to pick things up. (like the tag header option for the Varnish client)

does that help?

andrerom:
This will also have to be dealt with:
`$this->addTags(explode(',', $response->headers->get($this->getTagsHeaderName())));`

dbu:
oh, indeed. that one is for merging the tags, right?

andrerom:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/Http/SymfonyResponseTagger.php#L29-L42

dbu:
I tagged 2.2.0 of the bundle without this, because the Symfony 4 support has been waiting for a release for too long already. I have no problems tagging a 2.3.0 in a week or whenever we have the xkey support ready.

andrerom:
Good to know @dbu. This is up for grabs if anyone what to add the config for this btw, instructions above should give enough clues to it.

dbu:
fixed in #464

PR #451, opened 2018-05-10
Xkey Support

PeteLawrence:
This PR adds support for tagging and invalidating items using Varnish's xkey mode. (#431)

Enabling xkey based tagging and invalidating is done by setting the `fos_http_cache.proxy_client.varnish.tag_mode` setting to `purgekeys`.  Doing so will update the tag_handler to use the `xkey` response header, and to separate tags using a space instead of a comma.  It will also pass the `tag_mode` setting to the FOS http-cache library, which takes care of sending the correct response headers on invalidation requests.

Setting a custom `tags_header` setting is also supported (for switching between `xkey-softpurge` and `xkey-purge`, or a custom one).

If these changes all make sense I'll update the documentation too.

### DIFF
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -29,6 +29,7 @@ varnish
         proxy_client:
             varnish:
                 tags_header: My-Cache-Tags
+                tag_mode: ban
                 header_length: 1234
                 default_ban_headers:
                     Foo: Bar
@@ -37,13 +38,6 @@ varnish
                         - 123.123.123.1:6060
                         - 123.123.123.2
                     base_url: yourwebsite.com
-
-``tags_header``
-"""""""""""""""
-
-**type**: ``string`` **default**: ``X-Cache-Tags``
-
-Header for sending tag invalidation requests to Varnish.
 
 ``header_length``
 """""""""""""""""
@@ -97,6 +91,26 @@ URL may contain a path. If you access your web application on a port other than
 
     Double-check ``base_url``, for if it is mistyped, no content will be
     invalidated.
+
+``tag_mode``
+"""""""""""""""""
+
+**type**: ``string`` **default**: ``ban``
+
+Select whether to invalidate tags using the xkey module or with ban requests. Supported modes: ``ban`` and ``purgekeys``.
+
+When using ``purgekeys`` the bundle will default to using soft purges.  If you do not want to use soft purge (either because your varnish modules version is too old to support it or because it does not fit your scenario), additionally set the tags_header option to xkey-purge instead of the default xkey-softpurge.
+
+.. note::
+
+    To use the purgekeys method, you need the `xkey vmod <https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst>`_ enabled and VCL to handle xkey invalidation requests as explained in the :ref:`FOSHttpCache library docs on xkey support <foshttpcache:varnish_tagging>`.
+
+``tags_header``
+"""""""""""""""
+
+**type**: ``string`` **default**: ``X-Cache-Tags`` if ``tag_mode`` is ``ban``, otherwise ``xkey-softpurge``
+
+Header for sending tag invalidation requests to Varnish.
 
 See the :ref:`FOSHttpCache library docs <foshttpcache:varnish configuration>`
 on how to configure Varnish.
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -134,6 +134,48 @@ class Configuration implements ConfigurationInterface
 
                     throw new InvalidConfigurationException('To enable the user context logout handler, you need to configure a ban capable proxy_client.');
                 })
+            ->end()
+            //Set the correct default tags header for the varnish client, depending on whether we use BAN or xkey
+            ->validate()
+                ->ifTrue(
+                    function ($v) {
+                        return
+                            array_key_exists('proxy_client', $v)
+                            && array_key_exists('varnish', $v['proxy_client'])
+                        ;
+                    }
+                )
+                ->then(function ($v) {
+                    // Overide the header name on invalidation requests
+                    if (empty($v['proxy_client']['varnish']['tags_header'])) {
+                        $v['proxy_client']['varnish']['tags_header'] = (Varnish::TAG_XKEY === $v['proxy_client']['varnish']['tag_mode'])
+                            ? Varnish::DEFAULT_HTTP_HEADER_CACHE_XKEY
+                            : Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS;
+                    }
+
+                    return $v;
+                })
+            ->end()
+            //Set the correct default response header and separator for the tagger, depending on whether we use BAN or xkey
+            ->validate()
+                ->always()
+                ->then(function ($v) {
+                    // Set a default response header name if none is provided in the config file
+                    if (empty($v['tags']['response_header'])) {
+                        $v['tags']['response_header'] = TagHeaderFormatter::DEFAULT_HEADER_NAME;
+                    }
+
+                    // If using Xkey override the response header name and separator
+                    if (array_key_exists('proxy_client', $v)
+                        && array_key_exists('varnish', $v['proxy_client'])
+                        && Varnish::TAG_XKEY === $v['proxy_client']['varnish']['tag_mode']) {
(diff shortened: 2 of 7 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -398,14 +398,27 @@ private function loadVarnish(ContainerBuilder $container, XmlFileLoader $loader,
     {
         $this->createHttpDispatcherDefinition($container, $config['http'], $this->getAlias().'.proxy_client.varnish.http_dispatcher');
         $options = [
-            'tags_header' => $config['tags_header'],
+            'tag_mode' => $config['tag_mode'],
         ];
+
+        // If a 'tags_header' setting has been provided use that. Otherwise the FOS http-cache library automatically selects the header to use based on the 'tag_mode' setting
+        if (isset($config['tags_header'])) {
```
dbu: i prefer `array_key_exists` over `isset` because it will notice the problem when `$config` is not defined (e.g. after refactoring). could you change to that?

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -465,8 +478,23 @@ private function loadCacheTagging(ContainerBuilder $container, XmlFileLoader $lo
...
-        $container->setParameter($this->getAlias().'.tag_handler.response_header', $config['response_header']);
+
         $container->setParameter($this->getAlias().'.tag_handler.strict', $config['strict']);
+
+        // If no response_header parameter has been already been set, use the one from the config
+        if (!$container->hasParameter($this->getAlias().'.tag_handler.response_header')) {
+            if (isset($config['response_header'])) {
+                $container->setParameter($this->getAlias().'.tag_handler.response_header', $config['response_header']);
+            } else {
+                $container->setParameter($this->getAlias().'.tag_handler.response_header', 'X-Cache-Tags');
```
dbu: this if/else could be shortened with PHP 7.1 syntax to `$container->setParameter($this->getAlias().'.tag_handler.response_header', $config['response_header'] ?? Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS);` (please keep using the constant for this header)

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -465,8 +478,23 @@ private function loadCacheTagging(ContainerBuilder $container, XmlFileLoader $lo
         }
 
         $container->setParameter($this->getAlias().'.compiler_pass.tag_annotations', true);
-        $container->setParameter($this->getAlias().'.tag_handler.response_header', $config['response_header']);
+
         $container->setParameter($this->getAlias().'.tag_handler.strict', $config['strict']);
+
+        // If no response_header parameter has been already been set, use the one from the config
+        if (!$container->hasParameter($this->getAlias().'.tag_handler.response_header')) {
```
dbu: this feels a bit magic. wdyt of instead having a validate step that sets the right response_header and separator? i think we should only change the headers if varnish is the only or the `default` client, and otherwise use the default Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS if the response_header is not set

src/DependencyInjection/Configuration.php
```diff
@@ -357,6 +356,10 @@ private function addProxyClientSection(ArrayNodeDefinition $rootNode)
                                     ->info('Map of additional headers to include in each ban request.')
                                     ->prototype('scalar')->end()
                                 ->end()
+                                ->enumNode('tag_mode')
```
dbu: can you please add an `info` node here to explain the tag modes a bit? something like "If you can enable the xkey module in Varnish, use the purgekeys mode for more efficient tag handling"

PeteLawrence:
Thanks for reviewing this so promptly.

Yes, I agree that it'd make more sense to handle all the defaults within the Configuration class.  I did have a shot at using the validate() function initially but was running into problems.  I shall give this another shot though.  Everything else makes sense and should be pretty straight forward.

PeteLawrence:
That should address all of your feedback now.  I'm away for a week now, so please don't think I've abandoned this if you don't hear anything!

src/DependencyInjection/Configuration.php
```diff
@@ -118,6 +118,28 @@ function ($v) {
...
+                        // Check whether 'purgekeys' has been enabled
+                        return
+                            array_key_exists('proxy_client', $v)
+                            && array_key_exists('varnish', $v['proxy_client'])
+                            && 'purgekeys' === $v['proxy_client']['varnish']['tag_mode']
+                        ;
+                    }
+                )->then(function ($v) {
+                    // Overide the header name on invalidation requests
+                    $v['proxy_client']['varnish']['tags_header'] = 'xkey-softpurge';
```
dbu: i think we need to do those overrides conditionally, only if nothing has been set explicitly. currently, xkey-purge would not work as expected. same with the values below.

we could remove the default values and the condition above, to always apply this validation and set default values based on whether its purgekeys or not. and if a value is already set, we would not overwrite it.

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -97,6 +93,20 @@ URL may contain a path. If you access your web application on a port other than
...
     invalidated.
 
+``tag_mode``
+"""""""""""""""""
+
+**type**: ``string`` **default**: ``ban``
+
+Select whether to invalidate tags using the xkey module or with ban requests. Supported modes: ``ban`` and ``purgekeys``.
+
+When using ``purgekeys`` the bundle will default to using soft purges.  If you do not want to use soft purge (either because your varnish modules version is too old to support it or because it does not fit your scenario), additionally set the tags_header option to xkey-purge instead of the default xkey-softpurge.
```
dbu: imho this should link to the xkey varnish documentation. can we add a section like this:
 
```
To use the purgekeys method, you need the `xkey vmod`_ enabled and VCL to handle xkey invalidation requests as explained in the :ref:`FOSHttpCache library docs on xkey support <foshttpcache:varnish_tagging>`.

File: Resources/doc/reference/configuration/proxy-client.rst [hunk 7]
```diff
@@ -37,11 +38,6 @@ varnish
                         - 123.123.123.2
                     base_url: yourwebsite.com
 
-``tags_header``
-"""""""""""""""
-
-**type**: ``string`` **default**: ``X-Cache-Tags``
-
 Header for sending tag invalidation requests to Varnish.
```
dbu: you moved the title and info, but not this line. please move it along to the tags_header section.

src/DependencyInjection/Configuration.php
```diff
@@ -118,6 +118,32 @@ function ($v) {
 
                     throw new InvalidConfigurationException('You need to configure a proxy_client for the logout_handler.');
                 })
+            ->end()
+            ->validate()
```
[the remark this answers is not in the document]
PeteLawrence: Yeah, using validate() doesn't quite seem right, however when using beforeNormalization(), the $v variable contains just the settings that have been defined in the config file, and not any of the default values, meaning I'd have to add in some extra code to assign the defaults.

So:
```php
$v['proxy_client']['varnish']['tags_header'] = (Varnish::TAG_XKEY === $v['proxy_client']['varnish']['tag_mode']) 
    ? Varnish::DEFAULT_HTTP_HEADER_CACHE_XKEY
    : Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS;
```

becomes:
```php
if (!array_key_exists('tag_mode', $v['proxy_client']['varnish'])) {
    $v['proxy_client']['varnish']['tag_mode'] = Varnish::TAG_BAN;
}

$v['proxy_client']['varnish']['tags_header'] = (Varnish::TAG_XKEY === $v['proxy_client']['varnish']['tag_mode'])
    ? Varnish::DEFAULT_HTTP_HEADER_CACHE_XKEY
    : Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS;
```

I'm not sure which approach is the lesser of two evils.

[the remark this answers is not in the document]
dbu: i see, thanks for the explanation. then lets keep using `validate`

XWB:
@PeteLawrence Can you rebase this PR with master branch?

dbu:
the extension needs to set the new container parameter. they are not copied automatically from the configuration, only those that we set in the extension class with $container->setParameter are later available.

XWB:
Travis passes \o/

dbu:
merged in #464 - thanks again @PeteLawrence !

PR #452, opened 2018-05-10
Documentation Fix: header -> response_header

PeteLawrence:
Fix documentation to reflect the fact the setting is called response_header and not header


dbu:
merged in 390b67953fb7143d0546cd0462a377961d184681

PR #461, opened 2018-06-13
fix user context logout handler auto enabled to not load it when the proxy client is not ban capable

### LINKED ISSUE #459: UserContext Caching with Symfony
The UserContext feature is documented to be[ working with Symfony Cache](http://foshttpcachebundle.readthedocs.io/en/latest/features/user-context.html) in the documentation.

However, this doesn't work anymore, because the [UserContextInvalidator](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/UserContextInvalidator.php#L39) requires the `BanCapable` interface, and this requirement is not matched for the Symfony ProxyClient. Therefore the application already fails when building the container.

I don't know what exactly is happening here, but either we get rid of the hard dependency on the `BanCapable` interface or the documentation should be changed in order to reflect that this feature is not working with Symfony Cache.

dbu:
fix #459

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -105,18 +105,34 @@ class Configuration implements ConfigurationInterface
             ->validate()
                 ->ifTrue(
                     function ($v) {
-                        return $v['user_context']['logout_handler']['enabled']
-                            && !isset($v['proxy_client']);
+                        return false !== $v['user_context']['logout_handler']['enabled'];
                     }
                 )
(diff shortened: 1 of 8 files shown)


dbu:
@danrot would you have a bit of time to test this fix?

danrot:
I feel like I have to, if you have taken the time to fix it :smiley: 

I tested it and it works :tada: 

Thanks a lot, I am really appreciating it!

PR #462, opened 2018-06-15
Autoconfigure custom context providers

XWB:
This would reduce some configuration. Right now we can autoconfigure almost all services, except our custom context provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Changelog
 
 ### Added
 
+* Autoconfigure support for custom context providers.
+
 * Autowiring support for the services in this bundle:
 
   - fos_http_cache.cache_manager => FOS\HttpCacheBundle\CacheManager
--- a/Resources/doc/reference/configuration/user-context.rst
+++ b/Resources/doc/reference/configuration/user-context.rst
@@ -231,6 +231,13 @@ Custom providers need to:
 * implement the ``FOS\HttpCache\UserContext\ContextProvider`` interface
 * be tagged with ``fos_http_cache.user_context_provider``.
 
+.. versionadded:: 2.4.0
+    Autoconfigure support has been added in version 2.4.0 and works from
+    Symfony 3.3 and above. If autoconfigure is enabled, your custom provider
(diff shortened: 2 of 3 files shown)


dbu:
do you need to autowire the context provider somewhere? i assumed that this is an internal service that nobody outside the bundle wants to use. what is your use case for this?

if we merge this, there should be a note in the changelog.

XWB:
Well, right now we have configure the customer context provider. So our configuration looks like:

```yml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    App\:
        resource: '../src/*'
        exclude: '../src/{Entity,Repository}'

    App\UserContext\UserCacheProvider:
        tags:
            - { name: fos_http_cache.user_context_provider, priority: 10 }
```

By moving this stuff to `registerForAutoconfiguration`, our configuration could be simplified to:

```yml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    App\:
        resource: '../src/*'
        exclude: '../src/{Entity,Repository}'
```

dbu:
oh wow, now i get it. i was not aware that this is possible. very cool, indeed! yes, lets have this.

can you please update the doc as well http://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/user-context.html#custom-context-providers to explain the new way without tag, and then have a `.. versionadded::2.2` box that explains this is only available from that version. and then keep the service with tag for old versions of symfony that do not have this feature. (i hope you get the idea: explain everything, but in a non-confusing way, and starting with the easy modern way ;-) )

XWB:
> oh wow, now i get it. i was not aware that this is possible. very cool, indeed! yes, lets have this.

You can read more about it at http://symfony.com/doc/current/service_container/tags.html#autoconfiguring-tags :) Symfony uses it to automatically tag controllers, listeners, twig extensions etc.

> and then have a .. versionadded::2.2 box

Just to be sure, the next release should be 2.4?

dbu:
time marches :-O yes, 2.4 is correct.

XWB:
@dbu A note has been added.

PR #466, opened 2018-06-26
Add SessionListener::onFinishRequest method to be inline with Symfony 3.4.12

### LINKED ISSUE #465: Symfony 3.4.12 issue with 1.x?
After update today to 3.4.12 we are seeing the following:
screen shot 2018-06-25 at 14 46 34

Not looked deeply at it yet, so for now just asking if someone has seen the same.

emodric:
Fixes #465 

Same issue is probably happening on master, so needs to be forward ported.

### DIFF
--- a/Tests/Unit/EventListener/SessionListenerTest.php
+++ b/Tests/Unit/EventListener/SessionListenerTest.php
@@ -47,6 +47,28 @@ class SessionListenerTest extends TestCase
         $listener->onKernelRequest($event);
     }
 
+    public function testOnFinishRequestRemainsUntouched()
+    {
+        $event = $this
+            ->getMockBuilder('Symfony\Component\HttpKernel\Event\FinishRequestEvent')
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $inner = $this
+            ->getMockBuilder('Symfony\Component\HttpKernel\EventListener\SessionListener')
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $inner
+            ->expects($this->once())
+            ->method('onFinishRequest')
+            ->with($event)
+        ;
+
+        $listener = $this->getListener($inner);
+        $listener->onFinishRequest($event);
+    }
+
     /**
      * @dataProvider onKernelResponseProvider
      */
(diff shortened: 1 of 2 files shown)


XWB:
Wouldn't it be easier to extend `AbstractSessionListener` and remove the `getSubscribedEvents` method?

emodric:
@XWB That wouldn't prevent the original Symfony event listener to set the response to private, one already extending the abstract listener.

andrerom:
+1, but it does not avoid this happening in the future.

Should we add a `__call()` instead with the if check here?? /cc @dbu @Toflar

Toflar:
I don't think this would reduce the risk of something like this happening. Quite the opposite, I think it could even cause potential errors to remain undiscovered. `__call()` would blindly forward to the decorated service but potentially it should do something different 😄

andrerom:
> __call() would blindly forward to the decorated service but potentially it should do something different 😄

Well it can check which methods `BaseSessionListener::getSubscribedEvents()` hints about and only forward those. But yeah, I'm not saying that as a solution is pretty, just that we have a potential recurring fatal problem here.

andrerom:
So maybe we go ahead with this as is for now, you have my +1 so if @Toflar is +1 as well it's up to you @dbu _(or other maintainers)_

Tests/Unit/EventListener/SessionListenerTest.php
```diff
@@ -47,6 +47,28 @@ public function testOnKernelRequestRemainsUntouched()
         $listener->onKernelRequest($event);
     }
 
+    public function testOnFinishRequestRemainsUntouched()
```
Toflar: The test should be skipped if the method does not exist in the core session listener. See https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/396800434

emodric: @Toflar Done!

emodric:
Thank you @Toflar and @andrerom for review!

dbu:
I changed the lowest build to 2.8.0 in other projects, to reduce the possibility space. Could be done here too.

Github doey not allow me to merge on the phone when tests fail. @ddeboer could your merge this and tag a patch release? Otherwise i can try to do it tonight.

dbu:
merged in #472

PR #468, opened 2018-06-28
[1.x][Composer] Update dev dependencies to supported versions 

andrerom:
In order to get travis to not run out of memory on php 5: Update dependencies to supported versions, on dev dependencies even pick latests patch versions within all existing supported versions.

Todo:
- [ ] Adjust .travis.yml to match composer changes, but wait for #471 to be merged first

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,19 +23,19 @@
     "require": {
         "php": "^5.4 || ^7.0",
         "friendsofsymfony/http-cache": "^1.4.5",
-        "symfony/framework-bundle": "^2.3||^3.0"
+        "symfony/framework-bundle": "^2.8||^3.4"
     },
     "require-dev": {
         "mockery/mockery": "0.9.*",
         "monolog/monolog": "*",
-        "sensio/framework-extra-bundle": "^2.3||^3.0",
-        "symfony/symfony": "^2.8||^3.0",
-        "symfony/phpunit-bridge": "^3.2",
-        "symfony/expression-language": "^2.4||^3.0",
-        "symfony/monolog-bundle": "^2.3||^3.0",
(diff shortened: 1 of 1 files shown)


XWB:
You can also increase the memory limit, like https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/.travis.yml#L35

andrerom:
True, but as amount of possible releases to resolve grow that might not be enough on PHP5 eventually again.

But that said, if I continue down the path here I'll need to remove tests for older symfony releases, so I realize I need some input from @dbu / @ddeboer here on what is preferred.

composer.json
```diff
@@ -23,19 +23,19 @@
     "require": {
         "php": "^5.4 || ^7.0",
         "friendsofsymfony/http-cache": "^1.4.5",
-        "symfony/framework-bundle": "^2.3||^3.0"
+        "symfony/framework-bundle": "^2.8||^3.4"
```
dbu: i don't think we should suddenly restrict dependencies of the maintenance version. instead, we can change the versions in the `.travis.yml` file to only install more specific versions, that usually helps with the out of memory problems.

dbu:
ah sorry, the session listener hackery is part of the 1.3 version so thats why you wanted to fix that. i'll try the .travis.yml approach quickly.

composer.json
andrerom: Probably safest. Then we can just adjust require-dev, it should be enough to fix it, on top of corresponding travis.yml changes.

dbu:
#471 was enough. i suggest we drop this.

although i notice that composer.json allows far more versions than we actually test. but whatever, this is legacy code.

andrerom:
Ok, well you know where to come back when the issue comes back then ;)

PR #470, opened 2018-06-28
Fix session listener

### LINKED ISSUE #469: Using the AbstractSessionListener to fetch events breaks fos cache since symfony 3.4.12
The method onFinishRequest is not available in the session listener.

Since patch 3.4.12 this method is added and also returned in the getSubscribedEvents array.
This version contains the old array.
https://github.com/symfony/http-kernel/blob/v3.4.11/EventListener/AbstractSessionListener.php

Now the following exception is thrown: 
Warning: call_user_func() expects parameter 1 to be a valid callback, class 'FOS\HttpCacheBundle\EventListener\SessionListener' does not have a method 'onFinishRequest'

Line containing the error:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/6b75eba45d8c87082c2f66a6e9882b5a1359f72b/src/EventListener/SessionListener.php#L86

**Suggested fix:**

Implement getSubscribedEvents in the listener, not sure if this breaks anythings with the priorities.

samvdb:
Fixes #469

### DIFF
--- a/src/EventListener/SessionListener.php
+++ b/src/EventListener/SessionListener.php
@@ -84,10 +84,10 @@ final class SessionListener implements EventSubscriberInterface
 
     public static function getSubscribedEvents(): array
     {
-        return array(
-            KernelEvents::REQUEST => array('onKernelRequest', 128),
+        return [
+            KernelEvents::REQUEST => ['onKernelRequest', 128],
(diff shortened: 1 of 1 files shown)


dbu:
see #466

PR #473, opened 2018-06-29
Update master from 1.3

dbu:


dbu:
damn, now this is failing :-(
i have no idea why its good in 1.3 but not in master. i am running out of time, have to leave now. if somebody has time to investigate, i would appreciate it.

at least, for those on version 1 of this bundle, there is now 1.3.16 which contains the fix.

andrerom:
These are failures on Symfony 4, might be due to the changes @nicolas-grekas
 did on forcing cache to be private in some cases, but I haven't followed it closely.

```
1) FOS\HttpCacheBundle\Tests\Functional\EventListener\CacheControlListenerTest::testIsCached
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'public'
+'max-age=0, must-revalidate, private'
/home/travis/build/FriendsOfSymfony/FOSHttpCacheBundle/tests/Functional/EventListener/CacheControlListenerTest.php:27
2) FOS\HttpCacheBundle\Tests\Functional\EventListener\CacheControlListenerTest::testNotCached
Failed asserting that 'max-age=0, must-revalidate, private' contains "no-cache".
/home/travis/build/FriendsOfSymfony/FOSHttpCacheBundle/tests/Functional/EventListener/CacheControlListenerTest.php:37

```

nicolas-grekas:
This issue should be fixed by https://github.com/symfony/symfony/pull/27736

XWB:
Should we wait for the next Symfony 4.1.2 release?

dbu:
Guess we should mark the bogus symfony release as conflicting

Karel <notifications@github.com> schrieb am Di., 3. Juli 2018, 11:04:

dbu:
@nicolas-grekas this also does not work with 4.1.0. could it be that there is a bug with the NO_AUTO_CACHE_CONTROL_HEADER ?

nicolas-grekas:
I don't know really, it didn't change since it's introduction to me. Can you spot why it doesn't work?

dbu:
looking further, i see that other tests using the magic header do work. 

on the failing test, we do not set the header, but there is no cookies or anything, so overwriting the cache control should not happen: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/tests/Functional/EventListener/CacheControlListenerTest.php (and the path `/cached/42` has no firewall configured either)

dbu:
the [4.1 build](https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/400939731) with symfony/routing 4.1.x-dev (73770bf) fails. however, the [stability=dev build](https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/jobs/400939735) does NOT fail, even though it ends up with the same version of routing. now i am confused - are there other recent issues in symfony that could influence this?

it should not matter, but i checked that the fos/http-cache version is exactly the same on both builds.

XWB:
@dbu Symfony 4.1.2 has been released, perhaps you can give it another try?

dbu:
looks like its not even this MR but a general regression of some kind. https://travis-ci.org/FriendsOfSymfony/FOSHttpCacheBundle/builds/393516882 is master and was green a month ago, now i restarted it it has failures.

nicolas-grekas:
@dbu do you have more info? Maybe open an issue on Symfony if you have a reproducer and you found a suspicious behavior?

XWB:
Fixed in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/476

PR #483, opened 2018-09-27
Added support for the new MaxHeaderValueLengthFormatter

Toflar:
Let's start Hacktoberfest 2018 a bit earlier 😎 

Here's the support for https://github.com/FriendsOfSymfony/FOSHttpCache/pull/424. Although I found a very stupid issue in the library itself so that needs to be merged first (https://github.com/FriendsOfSymfony/FOSHttpCache/pull/426).


dbu:
cool, thanks! i merged and tagged the bugfix. it seems like the method is not used in the tests here, as they did not fail.

but oh yay, code rot. the failures are about lowest version build. i guess they also happen in master now.

composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "friendsofsymfony/http-cache": "^2.3",
+        "friendsofsymfony/http-cache": "^2.5",
```
dbu: ah, actually this is causing various problems with the lowest version build. 

lets bump the branch alias to 2.5 and increase version numbers as needed.

Toflar: I might have misunderstood here but I bumped it and it still fails :)

dbu: that was the start, now we have to get rid of old versions of things that we no longer can support. FOSHttpCache stopped supporting symfony 2.x in 2.5. we have to adjust here as well, and can remove the lts 2 build from the matrix.

Toflar: Ah true, wouldn't it be better to update master then and I rebase? Otherwise this PR includes a lot of actually unrelated stuff :)

dbu:
can you please rebase on master to see if this is now good?

Toflar:
Here we go 🎉

PR #484, opened 2018-09-28
Updated the bundle to use http-cache 2.5

Toflar:


dbu:
we need to drop the test with lts 2: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/19e8057bf0f3abe67ecb9cd2777895b590196692/.travis.yml#L51

the lowest version is probably that the monolog-bridge version 2.7 gets installed, according to the composer output, while we have the http-kernel at 3.4 and it does not have the logger interface. maybe we need to mark a conflict with monolog-brige < 3.0?

Toflar:
Done but still... :)

dbu:
`Downgrading symfony/monolog-bridge (v4.1.4 => v2.7.0): Loading from cache`

> the lowest version is probably that the monolog-bridge version 2.7 gets installed, according to the composer output, while we have the http-kernel at 3.4 and it does not have the logger interface. maybe we need to mark a conflict with monolog-brige < 3.0?

composer.json
```diff
@@ -52,7 +52,8 @@
         "symfony/console": "To send invalidation requests from the command line"
     },
     "conflict": {
-        "twig/twig": "<1.12.0"
+        "twig/twig": "<1.12.0",
+        "symfony/monolog-bridge": "<3.0"
```
dbu: can we use 3.4 here to be in sync with the other symfony components?

dbu:
thanks!

PR #485, opened 2018-10-01
Fix missing NoAutoCacheControl when vary on contex hash is set from Response

andrerom:
This moves logic to set NoAutoCacheControl so it is always executed if
vary header exists, now also when that is the case from response already.

Also adds unit test coverage for this feature.


dbu:
i will tag a bugfix release now.

oh, btw: it would be great if you could add changelog entries for such pull requests. i will go and add changelog entries for the things already merged.

andrerom:
> oh, btw: it would be great if you could add changelog entries for such pull requests. i will go and add changelog entries for the things already merged.

ok, I'll do that next time ;)

PR #486, opened 2018-10-02
use tag for user context hash request invalidation

### LINKED ISSUE #482: Cached objects are getting banned for every session
Hi all,

I am using Varnish 4.1.10 with Symfony 3.4.13 and PHP 7.1.16 in backend application, I have FOSHttpCacheBundle ^2.4@dev in my composer.json. I have configured user context based caching using **purgekeys** tag mode. 

I have used user role from our application to identify the user context, which makes one cached resource per role. 
 
As per my understanding is when I open a page as User1 for the first time, the page gets cached and returned from Varnish for the subsequent hits, this is working as expected as long as I am on same session, but the moment I logout and login, my immediate request is not returned from cache and served from a backend fetch, we can determine this with the “Age” response header.

After seeing the logs in varnish, I could see couple of BAN requests which were actually banning the content from being served from cache. And for the URL I hit, I see a cache miss saying **ExpBan nnn banned lookup**. I tried commenting out the code from fos_ban_recv subroutine from fos_ban.vcl and now I see cached pages served from varnish even after multiple sessions (logout/login). I do understand that these changes will impact the BAN functionality,  but for now I just tried.

My configurations in config.yml
```
fos_http_cache:
  user_context:
    enabled: true
  cache_manager:
    enabled: true
  tags:
    expression_language: axon.expression_language
    enabled: true
    rules:
      -
        match:
          path: myPage
        tag_expressions: ["'myPage-'~id"]
  proxy_client:
    varnish:
      tags_header: xkey-purge
      tag_mode: purgekeys
      header_length: 7500
      http:
        servers:
          - %fos_varnish_http_server%
        base_url: %fos_varnish_base_uri%
  invalidation:
    enabled: true
```

Now what I don't understand is the triggering point of those BAN request. The request from the logs look like the below,

`
Begin          req 229407 rxreq
-   Timestamp      Start: 1537183847.830529 0.000000 0.000000
-   Timestamp      Req: 1537183847.830529 0.000000 0.000000
-   ReqMethod      BAN
-   ReqURL         /
-   ReqProtocol    HTTP/1.1
-   ReqHeader      X-Host: .*
-   ReqHeader      X-Url: .*
-   ReqHeader      X-Content-Type: .*
`

Could someone assist me with this on identifying the source of the BAN requests. 
Please let me know if you need any information/logs.

Thanks!

dbu:
replace https://github.com/FriendsOfSymfony/FOSHttpCache/pull/427
fix #482


dbu:
@ranjzith can you check if this would work for you? i think this solution is more elegant than the BAN based approach even if we would make it work correctly.

@andrerom @lolautruche can you also look at this? i think it should work just fine, and also make the symfony httpcache setup more secure.

ranjzith:
@dbu, thanks for adding me here. This is indeed a confident approach to invalidate user context response objects from caching server.

Just to acknowledge my understanding on the approach,
- Have all the user context hash responses tagged with user identifier header(session id).
- Send invalidation/purge requests for the session id tag using httpcache on logout handlers.
- No changes in the current VCLs in this case.

andrerom:
Looks ok from my side.

However one thing to double check here:
- Check if the hash tag leaks into main request on Symfony Proxy, as it's the same request that happen in some weird cases for me _(on similar case in eZ Platform which abstracts 1.x atm still)_, so I ended up not using [tagger service](https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/EventSubscriber/UserContextSubscriber.php#L50-L52) in this case.

dbu:
@ranjzith yes, exactly. tag handling is already implemented and with this solution there is no need to change anything on varnish to make this work.

@andrerom do you create a new session when permissions of a user change? you could also delete the tagged hash lookup in that case so the new hash is created in the next request. is there something general in symfony to know when permissions change that we could implement in the bundle? (i think not, i guess dynamic permissions is application specific, right?)

leaking the tag into the main request would be unfortunate. looking at what is happening:
 - we have the user context plugin [handle](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/f5da3a98960b91cfd60fc60782b6ae9150e92033/src/SymfonyCache/UserContextListener.php#L193) the hash lookup.
- in my understanding that should be a full request and its a [master request](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/EventListener/TagListener.php#L111) 
- *however* neither [here](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/Http/SymfonyResponseTagger.php#L42) nor [here](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ResponseTagger.php#L94) do we reset the tags. as this will be the same php instance, we will indeed mess up here. 

i fear this is a serious bug in the ResponseTagger - it will acumulate tags if used multiple times. do you agree that the correct behaviour would be for the ResponseTagger to empty the tag list after tagging one response with the tags?

andrerom:
>  i guess dynamic permissions is application specific, right?

yes

> i fear this is a serious bug in the ResponseTagger - it will acumulate tags if used multiple times. do you agree that the correct behaviour would be for the ResponseTagger to empty the tag list after tagging one response with the tags?


yes, it's a design issue.

I've been thinking a bit about that in regards to how this should be _IF_ we where to propose moving tagging to Symfony itself. And been thinking trying to tie it to the request, and pass on to response once that is created or something to guarantee there is no sharing across requests. However don't think we can do that here easily, so just clearing it every-time we write the final response would be a good improvement on what we have.

dbu:
i think tying it to the request could get complicated with subrequests - we want to only tag the actual response that is sent to the cache. symfony would not aggregate headers from different sub requests. i can't think of a sane situation where we have in-process master requests while handling a master request.

see https://github.com/FriendsOfSymfony/FOSHttpCache/pull/430

dbu:
this should now be ready.

@ranjzith can you check with this branch in your installation to make sure we really fix the problem with this?

@andrerom could you maybe verify in Ez if we do not mess up with tagging here?

andrerom:
> @andrerom could you maybe verify in Ez if we do not mess up with tagging here?

Code wise it looks ok. But as we are still abstracting 1.x, and my stab during summer at refactoring to use 2.x is non functional, so I have no way to test that.

ranjzith:
With the changes I am facing couple of deprecation issues (I have PHP 7 and Symfony3.4 setup), couldn't really say its due to the changes but when I switch back to 2.4 branch everything gets back to normal. Sorry, could not test the changes right now with the setup.

PR #491, opened 2018-10-16
clarify how to use the symfony kernel dispatcher

### LINKED ISSUE #490: Confused about the kernel dispatcher documentation
I'm not correctly understanding what the `index.php` should look like for the [**kerneldispatcher**](https://foshttpcachebundle.readthedocs.io/en/latest/features/symfony-http-cache.html#optimization-for-single-server-installations)  setup:

If I look at the linked documentation: https://foshttpcache.readthedocs.io/en/latest/proxy-clients.html#kerneldispatcher-for-single-server-installations

The following lines:

```php
// Tell the kernel about the cache
$kernel->setHttpCache($httpCache);

// Create the Symfony proxy client with KernelDispatcher
$kernelDispatcher = new KernelDispatcher($kernel);
$symfony = new Symfony($kernelDispatcher);
```

are confusing me.

Should it not be enough the following implementation:

```diff
// ...
$kernel = new Kernel(...);

+ if (SYMFONY_ENV !== 'dev') {
+     $kernel = new HttpCache($kernel);
+ }
```

If I look at the symony basic [index.php](https://github.com/symfony/demo/blob/master/public/index.php) there is no KernelDispatcher or Symfony. Also not clear if $httpCache->handle need to be called or `$kernel->handle` I'm really confused if this lines are really needed.

dbu:
choosing some duplication of doc from the library to make things more understandable and less confusing.

fix #490

### DIFF
--- a/Resources/doc/features/symfony-http-cache.rst
+++ b/Resources/doc/features/symfony-http-cache.rst
@@ -11,7 +11,7 @@ running an application on shared hosting for instance
 You can use features of this library with the Symfony ``HttpCache``. The basic
 concept is to use event listeners on the HttpCache class.
 
-.. warning::
+.. note::
 
     Symfony HttpCache support is currently limited to following features:
 
@@ -20,7 +20,7 @@ concept is to use event listeners on the HttpCache class.
     * Cache Tags
     * User Context
 
-    Generic ``BAN`` operations are not supported.
+    Ban operations are not supported.
 
 Event Dispatching HttpCache
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


Resources/doc/features/symfony-http-cache.rst
```diff
@@ -65,9 +67,64 @@ dispatcher to directly call the ``HttpCache`` rather than sending an actual
...
+    $kernel = new AppKernel($env, $debug);
+    $cacheKernel = new CacheKernel($kernel);
+    // Tell the kernel about the cache
+    $kernel->setHttpCache($cacheKernel);
+
+    if ('prod' === $env) {
+        $response = $cacheKernel->handle($request);
+    } else {
+        $response = $kernel->handle($request);
+    }
```
alexander-schranz: sorry one question does `terminate` need to be called on $kernel or $cacheKernel or does it make no difference?

dbu: good point! i took up your suggestion of creating the cache kernel inside the kernel, so much easier this way!

PR #492, opened 2018-10-16
make tag annotations optional inside tags to allow using them without FrameworkExtraBundle

### LINKED ISSUE #489: Hard dependency on sensio/framework-extra-bundle when using cache tagging
I'm not quite sure if I got this right, so I figured I'll add an issue first and get your opinion but the parameter `fos_http_cache.compiler_pass.tag_annotations` in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/Compiler/TagListenerPass.php#L28 is set to `true` as soon as you enable cache tagging (https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/FOSHttpCacheExtension.php#L108-L119).

Why is this?

I mean, what's wrong in having all the cache tagging logic/services (`cache_tagging.xml`) without support for annotations? I only add my tags using code.

dbu:
fix #489


dbu:
i don't know how contao exactly works, but be aware that if users write their own controllers, they might rely on tag annotations and need to be aware of the change so that they overwrite the configuration and add the framework extra bundle themselves.

dbu:
@Toflar may i ask you to review this one and check if it works as intended in contao?

Toflar:
That looks good, yes! Just the `false` case is not covered in tests now :)

PR #493, opened 2018-10-23
Added link to symfony httpcache docs for http_method_override

alexander-schranz:
Based on the symfony docs enable method override is enabled by default so we should also add Request:: enableHttpMethodParameterOverride() by default?

https://symfony.com/doc/current/reference/configuration/framework.html#http-method-override

### DIFF
--- a/Resources/doc/features/symfony-http-cache.rst
+++ b/Resources/doc/features/symfony-http-cache.rst
@@ -98,8 +98,7 @@ your kernel like this::
         }
     }
 
-Now you need to adjust your front controller to use that cache instance rather
-than creating one::
+Now you need to adjust your front controller (that you set up according to the `Symfony HttpCache documentation_`) to use that cache instance rather than creating one::
 
     // public/index.php
 
@@ -111,6 +110,8 @@ than creating one::
     if ('prod' === $env) {
         $kernel = $kernel->getHttpCache();
     }
+
+    // ...
 
 .. warning::
 


dbu:
this documentation only shows what to do different for using the in-process communication to invalidate cache. its no complete code example of how to set up the CacheKernel. we don't show the `Request::createFromGlobals()` line. therefore i think we should not explain this special case (though its weird that the config default is true, but you need to explicitly enable it in the front controller if you want to use it with the cache kernel).

in the intro to this chapter, we already link to https://symfony.com/doc/current/http_cache.html#symfony-reverse-proxy

i think what would make sense is to change "Now you need to adjust your front controller to use that cache instance rather than creating one:" to "Now you need to adjust your front controller (that you set up according to the `Symfony HttpCache documentation`_) to use that cache instance rather than creating one:"

dbu:
do you agree that that would help?

Resources/doc/features/symfony-http-cache.rst
```diff
@@ -26,7 +26,7 @@ Event Dispatching HttpCache
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You need to adjust your ``AppCache`` to support event handling and register the
-cache event listeners for the functionality you want to use.
+cache event listeners for the functionality you want to use.git st
```
dbu: can you please remove this bit?

alexander-schranz:
@dbu I aggree. Did change the text as you propose.

Resources/doc/features/symfony-http-cache.rst
alexander-schranz: fixed

dbu:
thank you!

PR #495, opened 2018-10-23
Fix readthedocs code highlighting

alexander-schranz:
This should hopelly fix the code highlighting in the docs:
based on: https://github.com/sphinx-doc/sphinx/issues/3215#issuecomment-368623135

### DIFF
--- a/Resources/doc/conf.py
+++ b/Resources/doc/conf.py
@@ -28,6 +28,7 @@ extensions = [
     'sphinx.ext.coverage',
     'sphinx.ext.todo',
     'sensio.sphinx.configurationblock',
+    'sphinxcontrib.phpdomain',
     'sphinxcontrib.spelling'
 ]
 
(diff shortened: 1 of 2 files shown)


dbu:
oh, have not noticed but we don't highlight anymore :-(

thanks, looks like you are probably right. can you please add the extension as e requirement to
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Resources/doc/requirements.txt as well, to make the build succeed?

dbu:
thanks a lot!

dbu:
hm, seems not to help. i can't see in https://readthedocs.org/projects/foshttpcachebundle/builds/ what is wrong. the phpdomain extension seems to get installed. do you want to ask the readthedocs people on twitter if they have an idea why we get no highlighting?

alexander-schranz:
locally it looks good for me but I used the current version of the read the docs theme maybe update the theme version?

PR #499, opened 2018-10-23
Fixed symfony httpcache documentation link

alexander-schranz:
sorry seems like I broke the link 🙈


dbu:
thanks. weird, i'd have thought the doc build would spot that and fail - we do build the doc with sphinx on travis-ci to spot problems...

alexander-schranz:
@dbu it seems sphinx itself handles ` as italic text so a missing or a false placed `_` will not error, but github tries to render a link which looks strange there when the reference not found.

PR #510, opened 2018-12-20
Invalidate context after switch user

### LINKED ISSUE #506: Invalidate context after switch user
Currently invalidate executed after logout https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/Security/Http/Logout/ContextInvalidationSessionLogoutHandler.php. 
I had a problem with feature https://symfony.com/doc/current/security/impersonating_user.html. Should we be invalidation cache after switching?

Example of what I mean

```php
<?php
namespace App\EventSubscriber;

use FOS\HttpCacheBundle\UserContextInvalidator;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\Security\Http\Event\SwitchUserEvent;
use Symfony\Component\Security\Http\SecurityEvents;

class SwitchUserSubscriber implements EventSubscriberInterface
{
    private $invalidator;

    public function __construct(UserContextInvalidator $invalidator)
    {
        $this->invalidator = $invalidator;
    }

    /**
     * {@inheritdoc}
     */
    public static function getSubscribedEvents(): array
    {
        return [
            SecurityEvents::SWITCH_USER => 'onSwitchUser',
        ];
    }

    public function onSwitchUser(SwitchUserEvent $event): void
    {
        $this->invalidator->invalidateContext($request->getSession()->getId());
    }
}
```

onEXHovia:
Closes #506 

I added some tests, but not sure that fully covered this feature. Let me know if need to add something. Thanks!

### DIFF
--- a/src/EventListener/SwitchUserListener.php
+++ b/src/EventListener/SwitchUserListener.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCacheBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCacheBundle\EventListener;
+
+use FOS\HttpCacheBundle\UserContextInvalidator;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Security\Http\Event\SwitchUserEvent;
+use Symfony\Component\Security\Http\SecurityEvents;
+
+class SwitchUserListener implements EventSubscriberInterface
+{
+    private $invalidator;
+
+    public function __construct(UserContextInvalidator $invalidator)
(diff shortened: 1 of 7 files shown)


src/EventListener/SwitchUserListener.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace FOS\HttpCacheBundle\EventListener;
+
+use FOS\HttpCacheBundle\UserContextInvalidator;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Security\Http\Event\SwitchUserEvent;
+use Symfony\Component\Security\Http\SecurityEvents;
+
+class SwitchUserListener implements EventSubscriberInterface
```
dbu: can you add a phpdoc with the explanation why we do this? (basically the TLDR of our discussion in the issue you reported)

dbu:
i added the changelog and phpdoc and merged manually. thanks again for the contribution!

PR #516, opened 2019-03-12
do not leak the Symfony-Session-NoAutoCacheControl header when the Symfony session system is not enabled.

### LINKED ISSUE #512: Exposed Symfony-Session-NoAutoCacheControl header when sessions are disabled
The `UserContextListener` sets the `Symfony-Session-NoAutoCacheControl` header. But if a project does not even use sessions and disabled it, then the `\Symfony\Component\HttpKernel\EventListener\SessionListener` is not active. And thus also does not remove this header again. So FOSHttpCacheBundle should not set this header when sessions are not enabled in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/eacce6e423619fd2133862e185afeabf681a4aab/src/EventListener/UserContextListener.php#L151 and https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/eacce6e423619fd2133862e185afeabf681a4aab/src/EventListener/UserContextListener.php#L215

Otherwise it exposes this header to the outside.

dbu:
fix #512

### DIFF
--- a/src/EventListener/UserContextListener.php
+++ b/src/EventListener/UserContextListener.php
@@ -60,6 +60,14 @@ class UserContextListener implements EventSubscriberInterface
     private $options;
 
     /**
+     * Whether the application has a session listener and therefore could
+     * require the AbstractSessionListener::NO_AUTO_CACHE_CONTROL_HEADER.
+     *
+     * @var bool
+     */
+    private $hasSessionListener;
+
+    /**
      * @var string
      */
     private $hash;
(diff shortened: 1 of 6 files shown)


src/EventListener/UserContextListener.php
```diff
@@ -146,7 +156,7 @@ public function onKernelRequest(GetResponseEvent $event)
             $response->setClientTtl($this->options['ttl']);
             $response->setVary($this->options['user_identifier_headers']);
             $response->setPublic();
-            if (4 <= Kernel::MAJOR_VERSION && 1 <= Kernel::MINOR_VERSION) {
+            if ($this->hasSessionListener && version_compare('4.1', Kernel::VERSION, '<=')) {
```
dbu: the old code would have been broken with symfony 5.0. `<=` seems to work with all php versions: https://3v4l.org/GCltV

AyrtonRicardo: The old code failed with 5.0, but works with 5.1: https://3v4l.org/PtroA

dbu: exactly. thanks for looking at this, reminds me that i should merge it.

PR #520, opened 2019-04-28
Explain invalidation applies to unsafe requests

ddeboer:
While this should be evident, it helps people to document this explicitly.


dbu:
thank you!

PR #529, opened 2019-10-10
Use ContainerAwareTrait instead of deprecated ContainerAwareCommand

emodric:
This gets rid of the deprecation message on Symfony 4.2.

I've copied the `getContainer` method from `ContainerAwareCommand` to the command since the intention is to remove the usage of the container anyway for v3 of the bundle.


dbu:
tagged 2.7.2 with this, thank you @emodric

emodric:
Thanks @dbu !

PR #530, opened 2019-11-26
Adds compatibility with Twig 3.0

jderusse:


src/Twig/CacheTagExtension.php
```diff
@@ -12,11 +12,13 @@
 namespace FOS\HttpCacheBundle\Twig;
 
 use FOS\HttpCache\ResponseTagger;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
```
dbu: please add something like this

```
if (!\class_exists(AbstractExtension::class)) {
    class_alias(\Twig_Extension::class, AbstractExtension::class);
}
if (!\class_exists(TwigFunction::class)) {
    class_alias(\Twig_SimpleFunction::class, TwigFunction::class);
}
```

jderusse: Those classes exist since 1.34

https://github.com/twigphp/Twig/blob/v1.34.0/src/TwigFunction.php
https://github.com/twigphp/Twig/blob/v1.34.0/src/Extension/AbstractExtension.php

I think that travis tests with `prefer-lowest` would have turned red if it wasn't the case

dbu: hm, looks like, yes. weird, i think i saw some error about it. anyways, now its only the known sphinx problem so all good. thanks!

jderusse:
errors were related to a change in `$request->isMethodSafe` which is deprecated.
But the fix is not as easy as I thougth, and is not related to that PR, so I reverted it.

PR #532, opened 2019-11-27
Bump vendors

XWB:
Symfony 3.3 and 4.0 have been deprecated a long time ago.


dbu:
i think you worked on an outdated branch, master already is on 3.4 || 4, see https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/composer.json#L26

i agree that we don't necessarily need to support non-maintained versions, but i don't want to drop 4.0-4.2 support unless there turn out to be problems or work arounds required. as long as we support 3.4 anyways, we do need some legacy BC logic.

i am currently trying out a symfony 5 build, will push a PR soon.

XWB:
I understand. The reason for this change was the CVE alert for this repository:

image

To get rid of it, Symfony 4.0 must be dropped.

dbu:
ah, okay. i will bump to ^4.1.12 then. #533 is about to get ready - if you have some time to review the :puke: i had to do to run with both symfony 4 and 5, please do ;-)

PR #533, opened 2019-11-27
symfony 5 compatibility

dbu:
fix #532

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,16 +28,15 @@ matrix:
   include:
     # Minimum supported dependencies with the latest and oldest PHP version
 
-    - php: 7.1
-      env:
-        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
     - php: 7.2
       env:
-        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="symfony/validator:^3.4.4 symfony/templating:^3.4.4"
+    - php: 7.3
+      env:
+        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="symfony/validator:^3.4.4 symfony/templating:^3.4.4"
 
     # Test the latest stable release
 
-    - php: 7.1
     - php: 7.2
     - php: 7.3
       env:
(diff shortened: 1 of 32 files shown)


src/EventListener/CacheControlListener.php
```diff
@@ -16,8 +16,16 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\HttpKernel\KernelEvents;
 
+if (Kernel::VERSION >= 5) {
```
jderusse: beware, version is a string, what's about using `kernel::MAJOR_VERSION >= 5` or `kernel::VERSION_ID >= 50000`?

dbu: ups, yes i wanted to use MAJOR_VERSION here, thanks!

dbu:
hm, does anyone have an idea what is going on with the lowest version build here? why is the class loader not finding classes? i verified that the 3.4.5 file exists and contains the expected class (even bumped from 3.4.4 to be sure its not a cached broken archive or something)

jderusse:
@dbu because `UserPasswordValidator` extends `Symfony\Component\Validator\ConstraintValidator` which does not exists (the component is not installed)

dbu:
uh, that is rather implicit. thanks for pointing it out. that was the problem, had to add some dependencies. weird, that. security-bundle says it depends on those, not sure why they did not install.

do you think its ok to merge like this? i don't want to add them to `require` as thats only true for old versions, and the whole thing is require-dev only anyways...

jderusse:
not in 3.4 branch which is installed with prefered-lowest https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bundle/SecurityBundle/composer.json

jderusse:
not in 3.4 branch which is installed with prefered-lowest https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bundle/SecurityBundle/composer.json

.travis.yml
```diff
@@ -60,7 +62,8 @@ matrix:
 before_install:
   - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
   - if ! [ -z "$STABILITY" ]; then composer config minimum-stability ${STABILITY}; fi;
-  - if ! [ -v "$DEPENDENCIES" ]; then composer require --no-update ${DEPENDENCIES}; fi;
+  - if ! [ -z "$DEPENDENCIES" ]; then composer require --no-update ${DEPENDENCIES}; fi;
```
jderusse: sounds good to me

dbu:
yeah i was looking at require-dev at https://packagist.org/packages/symfony/security-bundle#v3.4.4 :man_facepalming:

PR #534, opened 2019-11-27
Fix deprecation with isMethodSafe

jderusse:


src/Http/RequestMatcher/UnsafeRequestMatcher.php
```diff
@@ -18,6 +18,14 @@ class UnsafeRequestMatcher implements RequestMatcherInterface
 {
     public function matches(Request $request)
     {
-        return !$request->isMethodSafe(false);
+        // hack needed for compatibility with SF 3.4 => 4.3
+        // sf 3.4 => isMethodSafe(false)
+        // sf 4.3 => isMethodSafe() or isMethodSafe(false)
+        // sf 4.4 => isMethodSafe()
+        if (method_exists($request, 'getPreferredFormat')) {
```
jderusse: Do you prefer using `Kernel::VERSION_ID` ? given this method `getPreferredFormat` is not related to the tested behavior?

dbu: that actually would be more explict. you could do `if (version_compare(Kernel::VERSION, '4.3.0', '>=')) {`

jderusse:
PR's ready (failing test is related to Documentation)

dbu:
great, thanks!

PR #539, opened 2020-03-26
move license file to root for current best practice

Tobion:


dbu:
thanks, was not aware that that changed!

PR #543, opened 2020-06-19
[Hash] Improve user hash change detection during request by comparing at the end

andrerom:
Regarding discussion on user hash change detection, this is what I had in mind:
- Makes sure we generate hash to compare with on response => In case hash changed during current request
- Given we only generate hash if `! isAnonymous()` as a performance tweak => Still keep track of if user was logged in at request time, in case that changed in the meantime

TODO:
- [x] Adapt tests _(getting a "endTest() : void must be compatible with ..", so we might need to adjust something here due to new mockery / phpunit version)_

### DIFF
--- a/src/EventListener/UserContextListener.php
+++ b/src/EventListener/UserContextListener.php
@@ -78,9 +78,9 @@ class UserContextListener implements EventSubscriberInterface
     private $hasSessionListener;
 
     /**
-     * @var string
+     * @var bool
      */
-    private $hash;
+    private $wasAnonymous;
 
     /**
      * Used to exclude anonymous requests (no authentication nor session) from user hash sanity check.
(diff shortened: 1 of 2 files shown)


andrerom:
> we would not have detected the hash change?

I think so yes, but besides some random report I don't know for sure if this is the cause or not, it just popped up when auditing the misc code involved as potential problem area.


> are we missing (unit) tests for some scenario?

probably, if you are ok with this change overall I'll try to find some time for tests in the next weeks

dbu:
yes, i do like the change and think it is cleaner than what we had. glad if you can look into a bit of testing to be sure we don't introduce a regression.

andrerom:
On hold for #545

dbu:
the build in master is fixed, can you please rebase this branch?

andrerom:
Done, all tests green 🟢 

Failure is on something not rated to the changes here:
> Exception occurred:
  File "/usr/lib/python2.7/dist-packages/sphinx/locale/__init__.py", line 195, in _
    return translators['sphinx'].ugettext(message)
KeyError: 'sphinx'

dbu:
yep, that failure is known. its the doc build not working. i tried to fix it in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/450 but never found a solution unfortunately.

dbu:
thanks a lot!

PR #544, opened 2020-07-22
Add invalidate all command

alexander-schranz:
It would be great to have also a command to invalidate the whole cache to call it in a deployment script.


src/Command/InvalidateDomainCommand.php
```diff
@@ -0,0 +1,86 @@
...
+        $this
+            ->setName(static::$defaultName) // BC with 2.8
+            ->setDescription('Invalidate a domain on all configured caching proxies')
+            ->addArgument(
+                'domains',
+                InputArgument::IS_ARRAY | InputArgument::REQUIRED,
+                'URL domains you want to invalidate, you can specify any number of domains'
+            )
+            ->setHelp(<<<'EOF'
+The <info>%command.name%</info> command invalidates a list of paths on the configured caching proxies.
```
dbu, line 59: this needs to be adjusted

src/Command/InvalidateDomainCommand.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(static::$defaultName) // BC with 2.8
+            ->setDescription('Invalidate a domain on all configured caching proxies')
+            ->addArgument(
+                'domains',
+                InputArgument::IS_ARRAY | InputArgument::REQUIRED,
+                'URL domains you want to invalidate, you can specify any number of domains'
```
dbu, line 56:
```suggestion
                'Domain names you want to invalidate. You can specify multiple domains separated by spaces.'
```

alexander-schranz:
@dbu should we implement the `ClearCapable` Interface into the Varnish Implementation and doing there the `BAN` with `.*`?

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A command to trigger cache invalidation by path from the command line.
+ *
+ * @author Alexander Schranz <alexander@sulu.io>
+ */
+class ClearCommand extends BaseInvalidateCommand
+{
+    use PathSanityCheck;
+
+    protected static $defaultName = 'fos:httpcache:invalidate';
```
dbu, line 27: i would prefer to call it fos:httpcache:clear to be explicit that this does not take options...

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+namespace FOS\HttpCacheBundle\Command;
+
+use FOS\HttpCacheBundle\CacheManager;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * A command to trigger cache invalidation by path from the command line.
```
dbu, line 19: this description is incorrect

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,69 @@
...
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->getCacheManager()->clearCache();
```
dbu, line 65: i suggest to do a check if the cache is clear capable here, then fallback to banning everything if the cache is ban capable, and finally report an explanation if neither is supported.

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(static::$defaultName) // BC with 2.8
+            ->setDescription('Invalidate the whole http cache.')
```
dbu, line 48:
```suggestion
            ->setDescription('Clear the HTTP cache.')
```

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+            ->setHelp(
+                <<<'EOF'
+The <info>%command.name%</info> command invalidates the whole cache in the configured caching proxies.
```
dbu, line 51:
```suggestion
The <info>%command.name%</info> command clears the whole cache or, if that is not supported, invalidates all cache entries in the configured caching proxies.
```

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $cacheManager = $this->getCacheManager();
+
+        if ($cacheManager->supports(CacheInvalidator::CLEAR)) {
+            $this->getCacheManager()->clearCache();
+        } elseif ($cacheManager->supports(CacheInvalidator::INVALIDATE)) {
+            $this->getCacheManager()->invalidateRegex('.*');
+        } else {
+            $output->writeln(
+                '<error>The configured http cache does not support "clear" or "invalidate".</error>'
```
dbu, line 74:
```suggestion
                '<error>The configured HTTP cache does not support "clear" or "invalidate".</error>'
```

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+
+namespace FOS\HttpCacheBundle\Command;
+
+use FOS\HttpCache\CacheInvalidator;
+use FOS\HttpCacheBundle\CacheManager;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * A command to trigger cache invalidation by path from the command line.
```
dbu, line 20:
```suggestion
 * A command to clear the whole cache from the command line.
```

alexander-schranz:
@dbu  seems like CacheInvalidator::CLEAR is only available since http-cache 2.6 should we increase the require http-cache version or do you prefer keep the min requirement and find another way?

composer.json
```diff
@@ -22,15 +22,15 @@
     ],
     "require": {
         "php": "^7.2",
-        "friendsofsymfony/http-cache": "^2.5.2",
+        "friendsofsymfony/http-cache": "^2.6",
```
alexander-schranz, line 25: CLEAR does not exit before 2.6: See https://travis-ci.org/github/FriendsOfSymfony/FOSHttpCacheBundle/jobs/714550545

.styleci.yml
```diff
@@ -2,7 +2,6 @@ preset: symfony
 
 enabled:
   - alpha_ordered_imports
-  - short_array_syntax
```
alexander-schranz, line 5: Else StyleCI Errors: https://github.styleci.io/analyses/EAVJNA?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

composer.json
dbu: yep, its fine to increase the version :+1:

.styleci.yml
dbu: thanks. seems that using the symfony preset is occasionally creating problems like this one.

composer.json
```diff
@@ -22,15 +22,15 @@
...
         "symfony/framework-bundle": "^3.4.26 || ^4.2.7 || ^5.0",
         "symfony/http-foundation": "^3.4.26 || ^4.2.7 || ^5.0",
         "symfony/http-kernel": "^3.4.26 || ^4.2.7 || ^5.0"
     },
     "require-dev": {
         "php-http/guzzle6-adapter": "^1.0 || ^2.0",
         "php-http/message": "^1.0 || ^2.0",
-        "mockery/mockery": "^1.0 || ^2.0",
+        "mockery/mockery": "^1.0",
```
dbu, line 33: why this change? are things broken with mockery 2 in master?

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+        parent::__construct($cacheManager);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(static::$defaultName) // BC with 2.8
```
dbu, line 47: this makes me wonder if we really need it - we don't support symfony 2.8 anymore...

src/Command/ClearCommand.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->getCacheManager()->invalidateRegex('.*');
+        } else {
+            $output->writeln(
+                '<error>The configured HTTP cache does not support "clear" or "invalidate".</error>'
+            );
+
+            return 1;
+        }
+
+        return 0;
```
dbu, line 80: i think this would be a bit clearer to read with early returns:

```
        if ($cacheManager->supports(CacheInvalidator::CLEAR)) {
            $this->getCacheManager()->clearCache();

            return 0;
        } 
        if ($cacheManager->supports(CacheInvalidator::INVALIDATE)) {
            $this->getCacheManager()->invalidateRegex('.*');

            return 0;
        } 

        $output->writeln(
            '<error>The configured HTTP cache does not support "clear" or "invalidate".</error>'
        );

        return 1;
```

(the code here is so simple that it does not matter that much, but i aim to have our code return early as a general principle)

phpunit.xml.dist
```diff
@@ -32,5 +32,6 @@
         <env name="KERNEL_DIR" value="./tests/Functional/Fixtures/app" />
         <env name="KERNEL_CLASS" value="AppKernel" />
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
+        <server name="SYMFONY_PHPUNIT_VERSION" value="6.5" />
```
dbu, line 35: i think this is breaking some tests `Call to undefined method FOS\HttpCacheBundle\Tests\Unit\Command\ClearCommandTest::assertStringContainsString()`

dbu: oh, actually those all come from the test you add, so the version of phpunit we use might simply not provide the StringContainsString assertion.

dbu:
thanks a lot! maybe it would make sense to do the build fix PR in a separate branch to keep it apart from the new command. would you mind splitting it out?

composer.json
alexander-schranz: @dbu There exist no mockery 2: https://packagist.org/packages/mockery/mockery

dbu: oh wow. i added that 2 years ago when allowing httplug 2 - seems i was over-motivated adding version 2 to things :man_facepalming: 

thanks for spotting!

dbu:
please rebase this now that the CI setup is fixed in master

alexander-schranz:
@dbu rebased and suggested changes done!

dbu (approved):
awesome, looks great!

one last thing: can you please add a note to the changelog (for next minor version) under the `Added` section?

alexander-schranz:
@dbu done!

dbu:
thanks a lot! tagged a release: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.9.0

alexander-schranz:
@dbu Awesome! Thank you for your fast reviews :)

PR #545, opened 2020-08-04
Fix CI Build and update PHPUnit

alexander-schranz:
Locally I run into the following problem:

> PHP Fatal error:  Declaration of FOS\HttpCacheBundle\Tests\Unit\CacheManagerTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in ~/Sulu/sulu-develop.localhost/vendor/friendsofsymfony/http-cache-bundle/tests/Unit/CacheManagerTest.php on line 28

> Fatal error: Declaration of FOS\HttpCacheBundle\Tests\Unit\CacheManagerTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in ~/Sulu/sulu-develop.localhost/vendor/friendsofsymfony/http-cache-bundle/tests/Unit/CacheManagerTest.php on line 28

This seems to be fixed by using:

```xml
<server name="SYMFONY_PHPUNIT_VERSION" value="6.5" />
```

Which seems to be set in the [.travis.yml](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/396eea2c79661fe591a80788de4d7ae671c76b49/.travis.yml#L20) also. I think I would remove it there so it works locally also.

But then I run into the following error:

> PHP Fatal error:  Declaration of Mockery\Adapter\Phpunit\TestListener::endTest(PHPUnit\Framework\Test $test, float $time): void must be compatible with PHPUnit\Framework\TestListener::endTest(PHPUnit\Framework\Test $test, $time) in /Users/alexanderschranz/Documents/Sulu/sulu20-develop.localhost/vendor/friendsofsymfony/http-cache-bundle/vendor/mockery/mockery/library/Mockery/Adapter/Phpunit/TestListener.php on line 39

So I decided to update to PHPUnit 8. Which seems to fix this error.

### DIFF
--- a/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
+++ b/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
@@ -41,9 +41,9 @@ final class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
         @trigger_error('Using the ContextInvalidationLogoutHandler is deprecated', E_USER_DEPRECATED);
 
         if (class_exists(LogoutEvent::class)) {
-            // This function should not longer be called in Symfony 5.1
+            // This class no longer works at all with Symfony 5.1, force usage of ContextInvalidationSessionLogoutHandler instead
             // See also: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/545#discussion_r465089219
-            throw new \LogicException(__CLASS__.'::'.__METHOD__.' is deprecated and should since Symfony 5.1 not longer be called.');
+            throw new \LogicException(__CLASS__.'::'.__METHOD__.' no longer works with Symfony 5.1. Remove fos_http_cache.user_context.logout_handler from your firewall configuration. See the changelog for version 2.2. for more information.');
         }
 
         $this->invalidator->invalidateContext($request->getSession()->getId());
--- a/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
+++ b/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
@@ -15,6 +15,7 @@ use FOS\HttpCacheBundle\UserContextInvalidator;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Http\Event\LogoutEvent;
 use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
 
 /**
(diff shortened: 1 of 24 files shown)


alexander-schranz:
@dbu any idea about the failing test?

alexander-schranz:
Seems like a regression in: https://github.com/symfony/symfony/pull/36243/files#r465083756

tests/Functional/Security/Http/Logout/ContextInvalidationLogoutHandlerTest.php
```diff
@@ -16,13 +16,19 @@
...
 
 class ContextInvalidationLogoutHandlerTest extends WebTestCase
 {
     use MockeryPHPUnitIntegration;
 
     public function testLogout()
     {
+        if (class_exists(LogoutEvent::class)) {
+            // @see https://github.com/symfony/symfony/pull/36243/files#r465083756
+            $this->markTestSkipped('This test does not work with Symfony 5.1.');
```
dbu, line 29: lets wait until tomorrow if we get a reaction on for https://github.com/symfony/symfony/pull/36243/files#r465083756 and then either remove this again, or also add a check to the actual implementation to make people realize the handler does not work with symfony 5.1

alexander-schranz: Are you sure you want to throw an exception when the class `ContextInvalidationLogoutHandlerTest` is used in Symfony 5.1? I'm little bit afraid doing that 🙈

dbu: i think its the correct thing to do. its not something we can silently work around, it simply does not work at all so devs need to be made aware if they did not check the changelog. better a clear error than a security issue that is silently sitting there.

alexander-schranz: Done in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/545/commits/853eaa0ea8bdcbf44769fc935f680dd843d189e0

src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
```diff
@@ -38,6 +39,13 @@ public function __construct(UserContextInvalidator $invalidator)
     public function logout(Request $request, Response $response, TokenInterface $token)
     {
         @trigger_error('Using the ContextInvalidationLogoutHandler is deprecated', E_USER_DEPRECATED);
+
+        if (class_exists(LogoutEvent::class)) {
+            // This function should not longer be called in Symfony 5.1
+            // See also: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/545#discussion_r465089219
+            throw new \LogicException(__CLASS__.'::'.__METHOD__.' is deprecated and should since Symfony 5.1 not longer be called.');
```
dbu, line 46:
```suggestion
            throw new \LogicException(__CLASS__.'::'.__METHOD__.' no longer works with Symfony 5.1. Remove fos_http_cache.user_context.logout_handler from your firewall configuration. See the changelog for version 2.2. for more information.');
```

src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
```diff
@@ -38,6 +39,13 @@ public function __construct(UserContextInvalidator $invalidator)
     public function logout(Request $request, Response $response, TokenInterface $token)
     {
         @trigger_error('Using the ContextInvalidationLogoutHandler is deprecated', E_USER_DEPRECATED);
+
+        if (class_exists(LogoutEvent::class)) {
+            // This function should not longer be called in Symfony 5.1
```
dbu, line 44:
```suggestion
            // This class no longer works at all with Symfony 5.1, force usage of ContextInvalidationSessionLogoutHandler instead
```

dbu (changes requested):
some wording changes for the exception, otherwise i am happy with it

dbu:
thanks a lot!

PR #548, opened 2020-11-27
[Bug] Flash messages are overwritten with redirects

ismailcherri:
## How to reproduce
1. Add a flash message on a controller
2. Return a `RedirectResponse` from that controller to another one
3. Add another flash message on the second controller

## What is expected
The `flashes` cookie should contain both messages

## What happens
When the second controller adds the flash message, the `FlashMessageListener` tries to get the cookie from the response header and can't find it, thus doesn't merge the old message with the new one in the newly formed cookie.

## Solution
Check if the request is a redirect and only set the cookie to the last response

### DIFF
--- a/src/EventListener/FlashMessageListener.php
+++ b/src/EventListener/FlashMessageListener.php
@@ -88,14 +88,20 @@ final class FlashMessageListener implements EventSubscriberInterface
             return;
         }
 
+        $response = $event->getResponse();
+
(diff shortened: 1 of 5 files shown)


dbu (approved):
this pull request is exemplary in quality: showing the error, then providing a fix. great job, thanks a lot!

dbu:
the styleci failure is unrelated, i fixed those on master now.

PR #550, opened 2020-12-02
disable doccheck and test with php 7.4

dbu:


dbu:
@ausi i wanted to start with disabling the doccheck and noticed that we don't test php 7.4

it seems either some edge case handling or some testing setup fails with 7.4 because it is more strict about array keys... can you start by looking into this? i bet that also fails with php 8...

ausi:
I will. Can I integrate these changes into my PHP 8.0 PR?

dbu:
yes sure, happy if you integrate them.

dbu:
fixed in #555

PR #552, opened 2020-12-03
PHP 8.0 Compatibility

dbu:
pushing #551 to this repo in the hope that travis-ci then sees it again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,28 +30,35 @@ matrix:
     - php: 7.2
       env:
         - COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="symfony/validator:^3.4.4 symfony/templating:^3.4.4"
-    - php: 7.3
+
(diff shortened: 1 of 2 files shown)


dbu:
oh, `Builds have been temporarily disabled for public repositories due to a negative credit balance.`
sigh, guess its time to move to github actions :-(

Yoann-TYT:
Hello @dbu :) 
Seems to be okay for merge ?

dbu:
the problem is that we don't get any travis-ci builds anymore because aparently it is "used up". it said it would replenish last week, but seems it did not. meanwhile the "plan page" only gives a tiny "There was an error" message.

so i think we need to rewrite the .travis-ci.yml to github actions to see if the tests build or not. while this looks nice and green, the phpunit tests are not executed at all, so i am not sure if it works. if somebody has time to rewrite the tests, awesome, otherwise i try to do it when i find time for it.

dbu:
fixed in #555

PR #554, opened 2020-12-11
Move from travis to GitHub action

Yoann-TYT:
## Move from travis to github actions. 

For the time being, I just play tests on PHP 7.2 and 7.3 with success


dbu:
thanks a lot for helping out here @Yoann-TYT !

the build on your repository looks good. any chance you can add the lowest version build and the builds with symfony 3 and 4, like we have in travis-ci? i guess we can leave the doccheck build out, it does not currently work anyways because somethingsomething with sphinx and the used python version

Yoann-TYT:
You're welcome 🙂

I'm checking it this week-end!

dbu:
thank you!

is there a way with github actions to avoid multiplying the builds? [actions](https://github.com/Yoann-TYT/FOSHttpCacheBundle/actions/runs/417763139) now does every php version with every symfony version, while [travis](https://travis-ci.org/github/FriendsOfSymfony/FOSHttpCacheBundle/builds/746309224) did every php version once, and every symfony version with the latest php version. (plus the --prefer-lowest build to make sure our lowest version declarations are correct)

dbu:
the syntax seems to be to add an `include` section inside the `matrix`. what do you think of only doing `include` within the `matrix`, like we did in travis.yml?

Yoann-TYT:
@dbu Done ;)
But i added the annotation "@runInSeparateProcess" on some functional tests to fix them

.github/workflows/php.yml
```diff
@@ -0,0 +1,64 @@
...
+  pull_request:
+    branches: [ master, /^\d+\.\d+$/, hotfix/move-to-github-action ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-version: ['7.2', '7.3']
+        composer-flag: ['--prefer-stable --prefer-lowest']
```
dbu, line 18: this should go away, we only want to have the 2 specific builds to prefer-lowest

.github/workflows/php.yml
```diff
@@ -0,0 +1,64 @@
...
+        dependency: ['symfony/validator:^3.4.4 symfony/templating:^3.4.4', 'symfony/flex', 'symfony/lts:^3']
```
dbu, line 19: i think this can also go away. to test the various php versions, we want to know if it works with default settings wihtout further restrictions. lts 3 is covered below, as is symfony/flex for symfony version 4.

dbu:
> But i added the annotation "@runInSeparateProcess" on some functional tests to fix them

funny that this changed, but fine for me. those being functional tests, i expect the added cost of starting separate processes to be cheap.

Yoann-TYT:
Seems that --prefer-lowest on PHP 7.4 fail for the functional tests about invalidateRoute and invalidatePath annotations ... 😢 
I will see that tomorrow

.github/workflows/php.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Check PHP Version
+        run: php -v
+      - name: Stability
+        run: composer config minimum-stability ${{ matrix.stability }}
+        if: ${{ matrix.stability }}
+      - name: Symfony version
+        run: composer require --no-update symfony/flex && composer config extra.symfony.require ${{ matrix.symfony-version}}
+        if: ${{ matrix.symfony-version }}
+      - name: Composer update
+        run: composer update --prefer-stable --prefer-lowest --prefer-dist --no-interaction
```
ddeboer, line 67: Did you intend to always prefer lowest? There’s no conditional here.

.github/workflows/php.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Composer validate
+        run: composer validate --strict --no-check-lock
+      - name: Install PHPUnit
+        run: php vendor/bin/simple-phpunit install
```
ddeboer, line 71: If I’m not mistaken, the PHPUnit bridge should itself install dependencies when running the tests for the first time.

Yoann-TYT: It's true :)

.github/workflows/php.yml
Yoann-TYT: Hi @ddeboer !
Yes, to avoid multiply test suites... do you think it's a bad practice and i should also test the latests ? 
Thanks for your review :)

ddeboer: Thank you for your PR! 😄

In my opinion, tests should also run against the latest versions of vendors, so having `--prefer-lowest` on all tests seems wrong. @dbu What do you think?

ddeboer (commented):
Feel free to remove `.travis.yml` in this PR, too.

.github/workflows/php.yml
dbu: i agree with @ddeboer . sorry if i was unclear with https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/554#discussion_r542225232

the idea of prefer lowest is to have 1 build with the lowest supported php version that installs the oldest allowed combination of packages, to check that the minimum requirements we have in composer.json work as advertised. (e.g. we could be using a class that has only been added in 1.3 of a dependency, but in composer.json we say `^1.2`. if we only build with the newest versions, we don't notice that mistake in the CI).
the rest of the builds however should use the newest version of dependencies, as that is the most likely way how the bundle will be running. 

i suggest to use the composer_flag variable to manage these options, but only set the flag in a matrix.include that uses the lowest php version we support.

.github/workflows/php.yml
```diff
@@ -0,0 +1,71 @@
...
+            stability: 'dev'
+          - php-version: '7.2'
+            symfony-version: '^4'
+            stability: ''
+          - php-version: '7.3'
+            symfony-version: '^4'
+            stability: ''
+          - php-version: '7.4'
+            symfony-version: '^4'
+            stability: ''
```
dbu, line 36: imho we don't need quite that many different builds. i think we want:

* every php version we support without symfony-version
* every symfony version we support with a suitable php version (aka symfony 4 with php 7.2 and symfony 5 with 7.3, 7.4 and 8.0)
* a lowest version build with php 7.2
* a build with stability dev and php 8.0

the goal is to cover all reasonable cases in as few builds as possible.

ddeboer: To reduce the number of builds further, feel free to remove PHP 7.2, which is [EOL anyway](https://www.php.net/supported-versions.php).

dbu: if we drop PHP 7.2 we need to also adjust composer.json to not allow installing the bundle on that php version.

.github/workflows/php.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - php-version: '7.3'
+            symfony-version: ''
+            stability: ''
+            composer-flag: ''
```
ddeboer, line 19: Just checking: is it really necessary to have these empty vars (`stability` and `composer-flag`) or can you leave them out?

ddeboer (approved):

.github/workflows/php.yml
Yoann-TYT: No, it seems to be unnecessary :)

.github/workflows/php.yml
Yoann-TYT: Thanks you :) It's done ;)

.github/workflows/php.yml
```diff
@@ -0,0 +1,54 @@
+name: CI FOSHttpCacheBundle
```
dbu, line 1:
```suggestion
name: CI
```

This name is used in the build badge in the readme. the simpler it is the better.

.github/workflows/php.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: composer config minimum-stability ${{ matrix.stability }}
+        if: ${{ matrix.stability }}
+      - name: Symfony version
+        run: composer require --no-update symfony/flex && composer config extra.symfony.require ${{ matrix.symfony-version}}
+        if: ${{ matrix.symfony-version }}
+      - name: Composer update
+        run: composer update ${{ matrix.composer-flag }} --prefer-dist --no-interaction
+      - name: Composer validate
+        run: composer validate --strict --no-check-lock
+      - name: Run the tests
```
dbu, line 53:
```suggestion
      - name: Run tests
```

dbu (approved):
thanks a lot!

i will merge on the cli to update changelog and readme as well.

dbu:
wrapped up in #555 - thank you very much!

PR #555, opened 2020-12-16
php 8 and github workflow

dbu:


ausi:
Thank you @dbu !

Is it possible to release 2.10.0 soon? ☺️

leofeyer:
A version 2.9.2 would also help. 😊

dbu:
i would love to get #556 in before tagging. if that did not happen by monday, i will tag for just the php 8 support.

as we increase the lowest php version, i prefer to do a minor version and not just a patch release. that leaves us the option of doing security patch releases for people on older php versions, if necessary.

leofeyer:
That absolutely makes sense. 👍🏻

dbu:
we managed to wrap it up, https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.10.0 is released

leofeyer:
Thank you! 🎉

PR #556, opened 2020-12-16
 Added support for php 8 attributes

Yoann-TYT:
##  Added support for php 8 attributes

The goal of this PR is to allow php 8 attributes instead doctrine annotations.

```php
#[Tag('foo')]
public function bar($id)
{
    //
}
```


tests/Unit/DependencyInjection/Compiler/HashGeneratorPassTest.php
```diff
@@ -49,11 +49,20 @@ public function testConfigNoContext()
         $config = $this->getBaseConfig();
         $this->extension->load([$config], $container);
         $this->userContextListenerPass->process($container);
-        if ($container->hasDefinition('service_container')) {
-            // symfony 3.3+
-            $this->assertCount(23, $container->getDefinitions());
+        if (method_exists(\ReflectionMethod::class, 'getAttributes')) {
+            if ($container->hasDefinition('service_container')) {
+                // symfony 3.3+
```
dbu, line 54: the minimum supported symfony version is 3.4 - could you remove the if statement and whole else block, while you are touching this?

tests/Unit/DependencyInjection/Compiler/HashGeneratorPassTest.php
```diff
@@ -49,11 +49,20 @@ public function testConfigNoContext()
         $config = $this->getBaseConfig();
         $this->extension->load([$config], $container);
         $this->userContextListenerPass->process($container);
-        if ($container->hasDefinition('service_container')) {
-            // symfony 3.3+
-            $this->assertCount(23, $container->getDefinitions());
+        if (method_exists(\ReflectionMethod::class, 'getAttributes')) {
```
dbu, line 52: i would prefer to check the php major version to be more explicit when this is the case.

tests/Functional/EventListener/TagListenerTest.php
```diff
@@ -47,6 +47,23 @@ public function testAnnotationTagsAreSet()
         $this->assertEquals('item-123', $response->headers->get('X-Cache-Tags'));
     }
 
+    /**
+     * @requires PHP 8.0
```
dbu, line 51: oh, i did not know that one. cool!

i was looking for some `$this->markSkipped` call...

src/Resources/config/catch_attributes.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="fos_http_cache.controller_resolver" class="Symfony\Bundle\FrameworkBundle\Controller\ControllerResolver">
```
dbu, line 7: what is the reason to define our own service for this? is the controller resolver not available from symfony?

src/EventListener/CatchAttributesListener.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+if (Kernel::MAJOR_VERSION >= 5) {
+    class_alias(RequestEvent::class, 'FOS\HttpCacheBundle\EventListener\AttributeRequestEvent');
+} else {
+    class_alias(GetResponseEvent::class, 'FOS\HttpCacheBundle\EventListener\AttributeRequestEvent');
+}
+
+class CatchAttributesListener implements EventSubscriberInterface
```
dbu, line 19: i am not too happy with the "Catch" name. what about `Php8AttributesListener`?

src/EventListener/CatchAttributesListener.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $this->controllerResolver = $controllerResolver;
+    }
+
+    public function onKernelRequest(AttributeRequestEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if (
+            method_exists(\ReflectionProperty::class, 'getAttributes') &&
```
dbu, line 36: should this listener be loaded at all when we are in php 7? maybe we could instead throw an exception in the constructor (and i would prefer to check the php version over method_exists here as well)

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -142,6 +142,10 @@ public function load(array $configs, ContainerBuilder $container)
 
             $loader->load('flash_message.xml');
         }
+
+        if (method_exists(\ReflectionProperty::class, 'getAttributes')) {
```
dbu, line 146: can we check php major version instead of method_exists?

src/EventListener/CatchAttributesListener.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+if (Kernel::MAJOR_VERSION >= 5) {
+    class_alias(RequestEvent::class, 'FOS\HttpCacheBundle\EventListener\AttributeRequestEvent');
+} else {
+    class_alias(GetResponseEvent::class, 'FOS\HttpCacheBundle\EventListener\AttributeRequestEvent');
+}
+
```
dbu, line 18: can you please add a phpdoc here that explains that this is about handling the php 8 attributes?

it has been a while that i looked into the codebase - how are the doctrine annotations provided into the request attributes? is that symfony itself doing it? does it not do that for php 8 annotations?

src/EventListener/CatchAttributesListener.php
```diff
@@ -0,0 +1,76 @@
...
+    public function onKernelRequest(AttributeRequestEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if (
+            method_exists(\ReflectionProperty::class, 'getAttributes') &&
+            $controller = $this->controllerResolver->getController($request)
+        ) {
+            $class = new \ReflectionClass($controller[0]);
+            $method = $class->getMethod($controller[1]);
```
dbu, line 40: according to the signature, `getController` returns any callable. some application might do different than class methods... to be on the safe side, i would rewrite the start of this method as:

```
$controller = $this->controllerResolver->getController($request);

if (!is_array($controller) || 2 !== count($controller)) {
    return;
}

$class = ...
```

dbu (changes requested):
thanks a lot for this pull request!

i am glad how you made it possible to integrate this into the existing annotation classes.

i added some nitpicks on the code.
once we have it refinded, this also needs an entry in the changelog and should be added in the documentation in the Resources/doc/ folder.

src/EventListener/CatchAttributesListener.php
Yoann-TYT: Yes, I'll add phpdoc for this class :) 

In Symfony, If we look the Routing component, we have a class named AnnotationClassLoader who load doctrine annotations or php8 attributes. They inject in the constructor Doctrine\Common\Annotations\Reader to fetch doctrine annotations.... And they use a check on PHP_VERSION_ID and not on method_exists to fetch php8 attributes :-D

Yoann-TYT:
Thanks for your review @dbu :) 

I took this into account in my last commit

dbu (approved):
looks great!

@ddeboer any input from your side or good to merge?

Resources/doc/reference/annotations.rst
```diff
@@ -28,6 +28,16 @@ Invalidate a path::
     {
     }
 
+From PHP 8, you can replace the doctrine annotations by PHP attributes::
+
+    use FOS\HttpCacheBundle\Configuration\InvalidatePath;
+
+    #[InvalidatePath('/articles')
+    #[InvalidatePath('/articles/latest')
```
ddeboer, line 36:
```suggestion
    #[InvalidatePath('/articles')]
    #[InvalidatePath('/articles/latest')]
```

Yoann-TYT: well seen !

dbu:
thanks a lot! i now released https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.10.0

Yoann-TYT:
@dbu  @ddeboer Thanks for your reviews :)

PR #557, opened 2020-12-22
Error in Php8AttributesListener

Yoann-TYT:
Sorry, I found a typo in my PR :-/ 

We need to use ControllerResolverInterface as type instead of Symfony\Bundle\FrameworkBundle\Controller\ControllerResolver


dbu:
thanks, i will tag a patch release

PR #559, opened 2021-01-26
array_merge_recursive rather than delay setting the flash messages cookie

### LINKED ISSUE #558: Empty flash_messages cookie in tests with v2.9.1
Hi,

when upgrading to version 2.9.1 my tests fail due to empty cookie jar of flash messages in tests.
With version 2.9.0 everything works just fine.
I have seen that version 2.9.1 should fix something with multiple redirects and flash messages, so I guess this has something to do with that.
Do I need to make any changes in my code / config?

dbu:
fix #558

### DIFF
--- a/src/EventListener/FlashMessageListener.php
+++ b/src/EventListener/FlashMessageListener.php
@@ -87,14 +87,6 @@ final class FlashMessageListener implements EventSubscriberInterface
             return;
         }
 
-        $response = $event->getResponse();
-
-        // If the response is a redirect, we should wait until the final response
-        // is reached
-        if ($response->isRedirect()) {
-            return;
-        }
-
         $flashBag = $this->session->getFlashBag();
         $flashes = $flashBag->all();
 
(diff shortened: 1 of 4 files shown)


dbu:
@ismailcherri @spointecker may I ask you two to review this pull request and tell if you see any new pitfalls i might be introducing here?

src/EventListener/FlashMessageListener.php
```diff
@@ -87,26 +87,28 @@ public function onKernelResponse(FlashMessageResponseEvent $event)
...
         if (isset($cookies[$host][$this->options['path']][$this->options['name']])) {
             $rawCookie = $cookies[$host][$this->options['path']][$this->options['name']]->getValue();
-            $flashes = array_merge($flashes, json_decode($rawCookie));
+            $flashes = array_merge_recursive($flashes, json_decode($rawCookie, true));
+        }
+
+        // Preserve existing flash message cookie from previous redirect if there was one.
+        // This covers multiple redirects where each redirect adds flash messages.
+        $request = $event->getRequest();
+        if ($request->cookies->has($this->options['name'])) {
```
dbu, line 109: before, we did not look at request cookies at all.

i was wondering if there is a security risk involved with this (depending on what weird things a dev might put into a flash message) but i guess if the attacker manages to inject a cookie for this domain, they can also do more direct attacks with javascript that do not involve looping through the backend.

spointecker:
LGTM. Thanks for the quick answer and fix, will test it with my app&tests as soon as it is released!

ismailcherri:
LGTM. Thanks for the fix.

dbu:
thanks for the quick feedback!

released https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.10.2 and https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.9.2

PR #561, opened 2021-01-26
docs: use Symfony HttpClient by default

dunglas:


dbu (approved):
yeah being a symfony bundle, i think it makes sense to use the symfony client.

and thanks for the upgrades on configuration locations

PR #565, opened 2021-08-29
fix build with symfony 5

dbu:


.github/workflows/php.yml
```diff
@@ -60,3 +60,8 @@ jobs:
         if: ${{ matrix.stability != 'dev' }}
       - name: Run tests
         run: php vendor/bin/simple-phpunit -v
+        if: ${{ matrix.stability != 'dev' }}
+      - name: Run tests allow to fail
+        run: php vendor/bin/simple-phpunit -v || true
```
dbu, line 65: github actions seem to lack a way to say the build should still be green on failure... https://github.com/actions/toolkit/issues/399

PR #571, opened 2021-12-05
Compatibility Symfony 6.0

bastnic:
No idea if it will pass CI, let's check it.


dbu:
we seem to have some kind of regression with a dependency: i just altered the build config to run all matrix entries even if one fails, and now we see the same error: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/actions/runs/1543828580 - but everything was fine one month ago: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/actions/runs/1433692815

do you have time to look into what might be causing the regression? do we end up installing something symfony 6 anyways, or some other dependency?

bastnic:
thanks for the follow up. I'ven't that much time with school closed all the time :s.

dbu:
thanks for having started it!

PR #573, opened 2022-01-25
Sf6

dbu:
continue #571

problems
* we used to set up things with the session we got from the container, but that is no longer possible with symfony 6, as the session is not in the container anymore. TokenStorage does not contain the token of the authenticated user. UserContextListenerTest both fail. hash is not as expected and therefore the UserContextListener removes caching because hash changes
* somehow twig is not loaded. no idea why, we register the bundle and configure it. TagListenerTest uses twig.

if anyone has ideas, please tell me in a comment ;-)


tests/Functional/EventListener/FlashMessageListenerTest.php
```diff
@@ -21,10 +23,12 @@ class FlashMessageListenerTest extends WebTestCase
     public function testFlashMessageCookieIsSet()
     {
         $client = static::createClient();
-        $session = static::$kernel->getContainer()->get('session');
 
         $client->request('GET', '/flash');
-        $this->assertFalse($session->isStarted());
+        if (Kernel::MAJOR_VERSION < 6) {
+            $session = static::$kernel->getContainer()->get('session');
+            $this->assertFalse($session->isStarted());
```
dbu, line 30: is there a way to check that no session is started in symfony 6? the upgrade instructions say the session is only available from the request stack, but the stack is empty at this point here.

src/Resources/config/user_context.xml
```diff
@@ -33,7 +33,7 @@
 
         <service id="fos_http_cache.user_context.logout_handler" class="FOS\HttpCacheBundle\Security\Http\Logout\ContextInvalidationLogoutHandler" public="false">
             <argument type="service" id="fos_http_cache.user_context_invalidator" />
-            <deprecated>The "%service_id%" service is deprecated since 2.2 and will be removed in 3.0.</deprecated>
+            <deprecated version="2.2" package="friendsofsymfony/http-cache-bundle">The "%service_id%" service is deprecated since 2.2 and will be removed in 3.0.</deprecated>
```
dbu, line 36: sometimes supporting multiple symfony versions is really hard :-(

these attributes are required with symfony 6 but forbidden in older symfony versions. should we split the configuration file? does it even make sense to keep providing the legacy logout handler in symfony 6 or does it not work anyways?

dbu: i split the file

tests/Functional/EventListener/SwitchUserListenerTest.php
```diff
@@ -28,12 +32,18 @@
...
+    private const SESSION_ID = 'test';
+    private $sessionName;
+    private static $overrideService = false;
 
     public function testSwitchUserCompatibility()
     {
         $client = static::createClient();
-        $session = $client->getContainer()->get('session');
-        $this->loginAsAdmin($client, $session);
+        $this->callInRequestContext($client, [$this, 'loginAsAdmin']);
```
dbu, line 45: @wouterj inspired by https://github.com/symfony/symfony/pull/41883 i tried this, but it seems to not work as expected, neither for symfony 5 not symfony 6.

do you see what i am doing wrong, by any chance?

src/EventListener/SessionListener.php
```diff
@@ -77,7 +77,7 @@ public function onKernelRequest(SessionRequestEvent $event)
 
     public function onKernelResponse(SessionResponseEvent $event)
     {
-        if (!$event->isMasterRequest()) {
+        if (Kernel::MAJOR_VERSION >= 6 ? !$event->isMainRequest() : !$event->isMasterRequest()) {
```
acrobat, line 80: I would suggest doing a method_exists check, this to avoid triggering deprecations on symfony 5.4

```
$mainRequest = method_exists($event, 'isMainRequest') ? $event->isMainRequest() : $event->isMasterRequest();
```

dbu: ah good point. we could also version compare >= 5.3

it seems cleaner to me to check the kernel version than doing the method_exists check. do you think that would make sense?

tests/Functional/EventListener/SwitchUserListenerTest.php
dbu: @acrobat do you by any chance see what i do wrong with the session manipulation in the refactoring?

src/EventListener/SessionListener.php
acrobat: In this case it wouldn't indeed make a difference as the event class used here is in the same package as the Kernel class, but most of the time when I need those checks I use the method_exists for 2 reasons.

* The kernel version might be different as the package version of that specific package (if it's not located in the http-kernel package ofcourse)
* Calling the deprecated method when the new one already exists (in this case 5.4 has the method also) will still trigger deprecations on 5.4 which will make the user think that he needs to resolve it still somehow.

tests/Functional/EventListener/SwitchUserListenerTest.php
acrobat: @dbu I will take a look at it!

acrobat:
@dbu I've created a PR to fix some of the issues that cause failing tests here. See #575

For authenticating users there is (since 5.1) as [loginUser method](https://symfony.com/doc/current/testing.html#logging-in-users-authentication) that can be used. So I've refactored the tests to use this method or polyfill it in case of older symfony versions

The missing twig service was actually caused by the fact that the full container was injected into the controller (instead the specific service locator required by newer versions of the symfony abstract controller) and the `twig` service being private by default since symfony 6. This caused it to be removed from the container and thus missing when running the tests. So I've updated the test controller service definition to fully use the autowire/autoregister fuctionality and make sure the controller receive the specific service locator

dbu:
symfony 6 error: `No password hasher has been configured`

src/EventListener/SessionListener.php
```diff
@@ -77,7 +77,8 @@ public function onKernelRequest(SessionRequestEvent $event)
 
     public function onKernelResponse(SessionResponseEvent $event)
     {
-        if (!$event->isMasterRequest()) {
+        // BC for symfony < 5.3
+        if (method_exists($event, 'isMainEvent') ? !$event->isMainRequest() : !$event->isMasterRequest()) {
```
acrobat, line 81:
```suggestion
        if (method_exists($event, 'isMainRequest') ? !$event->isMainRequest() : !$event->isMasterRequest()) {
```

dbu: oh my. thanks :facepalm: 

i am too tired :/

src/FOSHttpCacheBundle.php
```diff
@@ -30,17 +29,10 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new LoggerPass());
         $container->addCompilerPass(new TagListenerPass());
         $container->addCompilerPass(new HashGeneratorPass());
-        if (version_compare(Kernel::VERSION, '3.4', '>=')
-            && version_compare(Kernel::VERSION, '4.1', '<')
```
dbu, line 34: this was a regression we introduced at some point. the compiler pass is still relevant with newer symfony versions, as it tells the UserContextListener whether to set a magic header for the symfony core session listener.

dbu:
ok, almost there. to make things simpler, i ditched symfony 3.4 and 4.3 support (3.4 users can keep using an existing stable release, 4.3 is EOL anyways)

now the last problem i run into is that the workaround to get a defined session id with symfony 6 works fine with symfony 4 and 5, but not 6. the problem i have is that `SessionAuthenticationStrategy::onAuthentication` is triggered with each request after the first one i do in the test, even when i already have done a successfully authenticated request. the strategy is `migrate`, resulting in a new session id. i spent quite some time with xdebug trying to trace down why this happens, but i am not familiar enough with the symfony security system to tell if that is expected or not. the calling stack is

```
SessionAuthenticationStrategy.php:45, Symfony\Component\Security\Http\Session\SessionAuthenticationStrategy->onAuthentication()
SessionStrategyListener.php:46, Symfony\Component\Security\Http\EventListener\SessionStrategyListener->onSuccessfulLogin()
EventDispatcher.php:270, Symfony\Component\EventDispatcher\EventDispatcher::Symfony\Component\EventDispatcher\{closure:/home/david/liip/os/bundles/FOSHttpCacheBundle/vendor/symfony/event-dispatcher/EventDispatcher.php:265-271}()
EventDispatcher.php:230, Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
EventDispatcher.php:59, Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
AuthenticatorManager.php:248, Symfony\Component\Security\Http\Authentication\AuthenticatorManager->handleAuthenticationSuccess()
AuthenticatorManager.php:226, Symfony\Component\Security\Http\Authentication\AuthenticatorManager->executeAuthenticator()
AuthenticatorManager.php:160, Symfony\Component\Security\Http\Authentication\AuthenticatorManager->executeAuthenticators()
AuthenticatorManager.php:140, Symfony\Component\Security\Http\Authentication\AuthenticatorManager->authenticateRequest()
AuthenticatorManagerListener.php:40, Symfony\Component\Security\Http\Firewall\AuthenticatorManagerListener->authenticate()
TraceableAuthenticatorManagerListener.php:65, Symfony\Component\Security\Http\Authenticator\Debug\TraceableAuthenticatorManagerListener->authenticate()
WrappedLazyListener.php:49, Symfony\Bundle\SecurityBundle\Debug\WrappedLazyListener->authenticate()
AbstractListener.php:26, Symfony\Component\Security\Http\Firewall\AbstractListener->__invoke()
TraceableFirewallListener.php:73, Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener->callListeners()
Firewall.php:92, Symfony\Component\Security\Http\Firewall->onKernelRequest()
EventDispatcher.php:270, Symfony\Component\EventDispatcher\EventDispatcher::Symfony\Component\EventDispatcher\{closure:/home/david/liip/os/bundles/FOSHttpCacheBundle/vendor/symfony/event-dispatcher/EventDispatcher.php:265-271}()
EventDispatcher.php:230, Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
EventDispatcher.php:59, Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
HttpKernel.php:128, Symfony\Component\HttpKernel\HttpKernel->handleRaw()
HttpKernel.php:74, Symfony\Component\HttpKernel\HttpKernel->handle()
Kernel.php:202, Symfony\Component\HttpKernel\Kernel->handle()
HttpKernelBrowser.php:65, Symfony\Component\HttpKernel\HttpKernelBrowser->doRequest()
KernelBrowser.php:172, Symfony\Bundle\FrameworkBundle\KernelBrowser->doRequest()
AbstractBrowser.php:370, Symfony\Component\BrowserKit\AbstractBrowser->request()
UserContextListenerTest.php:38, FOS\HttpCacheBundle\Tests\Functional\EventListener\UserContextListenerTest->testHashLookup()
TestCase.php:1491, PHPUnit\Framework\TestCase->runTest()
TestCase.php:1111, PHPUnit\Framework\TestCase->runBare()
TestResult.php:703, PHPUnit\Framework\TestResult->run()
TestCase.php:839, PHPUnit\Framework\TestCase->run()
TestSuite.php:626, PHPUnit\Framework\TestSuite->run()
TestSuite.php:626, PHPUnit\Framework\TestSuite->run()
TestRunner.php:656, PHPUnit\TextUI\TestRunner->doRun()
Command.php:235, PHPUnit\TextUI\Command->run()
Command.php:194, PHPUnit\TextUI\Command::main()
phpunit:22, include()
simple-phpunit.php:430, require()
simple-phpunit:13, include()
simple-phpunit:107, {main}()
```

i am not sure if i do something wrong, if there is a bug with mock file session or something else.

but to my understanding, the session id between requests should normally be stable, when you have the session cookie.

i thought i found the issue to be #46206 but that one is talking about symfony 5, while here all tests succeed but the symfony 6 tests.

dbu:
i don't really care about a specific session name, but we do need to know that the right tag is set so that invalidation requests will work correctly when changing the permissions of a user. 

we could alternatively record the tag instead of directly asserting it, and then try to extract the session name from the data to compare. but i am worried if the sesion id somehow is supposed to keep changing in symfony 6, which would break the whole invalidation concept for the user context.

tests/Functional/EventListener/UserContextListenerTest.php
```diff
@@ -11,34 +11,41 @@
...
         ] : []);
-        /** @var SessionInterface $session */
-        $session = $client->getContainer()->get('session');
-        $session->setId('test');
-        $session->start();
+
+        $this->callInRequestContext($client, [$this, 'setSessionId']);
+
+        // remove basic auth to avoid restarting the session
+        $client->setServerParameters([]);
```
dbu, line 38: okay, this at least works around the issue. not sure if it is supposed to happen or not (session gets migrated on every request when basic auth headers are sent in request).

dbu: thanks. i asked wouter on symfony slack what the idea with basic auth & session is. (its a weird edge case anyways, for our tests i am happy with this solution)

dbu:
finally :sweat_smile:

acrobat:
👏 thanks for all the work 💪

tjveldhuizen:
Indeed, thanks a lot!

I've tested the master-branch locally, resulting in no errors. Unfortunately our test coverage is insufficient to be able to say it is 100% OK, but at least it's a good sign.
Are there other issues that need to be fixed before we can expect a new release?

dbu:
thanks for confirming. i am somewhat confident in the test suite we have - mostly afraid of integration issues in existing installations. i went ahead and tagged https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.12.0 - worst case we do a 2.12.1 if something is amiss

PR #575, opened 2022-04-18
Fixes for SF6 support

acrobat:
_This builds onto #573 and fixes some of the issues._

Each commit fixes a certain problem

- Authentication of a user
- Missing twig service
- Incorrect notation of controller::action which causes failing tests
- The anonymous token class was removed and the token storage will return null instead of the anonymous token


src/UserContext/RoleProvider.php
```diff
@@ -52,10 +54,15 @@ public function updateUserContext(UserContext $context)
...
         }
 
-        if (null === $token = $this->tokenStorage->getToken()) {
+        $token = $this->tokenStorage->getToken();
+        if (null === $token && class_exists(AnonymousToken::class)) {
             return;
         }
 
+        if (null === $token) {
+            $token = new NullToken();
```
dbu, line 63: what is the purpose of instantiating a NullToken here? if token storage does not give us that null token, should we not avoid updating the context?

i think we should add a comment why we do this, and one above to explain until what symfony version this will be the case.

tests/Functional/EventListener/SwitchUserListenerTest.php
```diff
@@ -61,47 +57,63 @@ public function testInvalidateContext()
         $mock = \Mockery::mock(Varnish::class);
         $mock->shouldReceive('invalidateTags')
             ->once()
-            ->with(['fos_http_cache_hashlookup-test']);
+            ->with(\Mockery::any())
```
dbu, line 60: i liked that this verifies that the correct tag is invalidated.

dbu (commented):
thanks a lot! this makes some good progress :+1:

src/UserContext/RoleProvider.php
acrobat: Yes I was not totally sure of this solutation as the AnonymousToken is gone and the token storage just returns null if your not logged in. But the null token is like a replacement for the AnonymousToken by I think it's only used by symfony in the voter system so that you actually always receive a token.

I did the fix like this because it would fix the tests and seemed logic to me. As previously the AnonymousToken would return an empty array of user roles that was used in the calculation of the hash, with the new behaviour this was skipped and I only could get it to work with the null token. So we might have to look into this if it's the correct fix.

tests/Functional/EventListener/SwitchUserListenerTest.php
acrobat: Yeah I know, I wasn't to happy about this change. The thing because we do a real login the session gets the standard random id which we can't predict. I don't know mockery well enough but a pattern match is something we can do here so we do check part of the string?

This didn't for me 🤔  `->with(\Mockery::pattern('/^fos_http_cache_hashlookup\-.*/'))`

alexander-schranz:
Also looking forward for symfony 6 support here. Anything I can help here with?

PR #576, opened 2022-05-05
Fix test cases for user context listener with different http basic authentication

alexander-schranz:
While debugging Symfony 6 test case. I did found out that the authentication does on the existing master branch not work like expected.

The test is actually sending `user` as username but actually if you debug and output: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/36159bbac615d0e4e804d1c43e9ef88123f6a82b/src/UserContext/RoleProvider.php#L60

It is just empty. That is because http_basic is not activated and so the user not logged in.

So the hash which was actually tested is incorrect for `user` and is the hash for not `user`. I addded a data provider so a test runs against different users.


dbu (approved):
phantastic, thanks a lot!

PR #579, opened 2022-05-10
php 8.1 : Fix deprecated

cboucaut:
Hi,

I use the projet with PHP 8.1 and I've found 2 deprecated.

### DIFF
--- a/src/Http/RequestMatcher/QuerystringRequestMatcher.php
+++ b/src/Http/RequestMatcher/QuerystringRequestMatcher.php
@@ -45,6 +45,6 @@ class QuerystringRequestMatcher extends SymfonyRequestMatcher
             return true;
         }
 
-        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString()));
+        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString() ?: ''));
     }
 }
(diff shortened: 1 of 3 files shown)


src/Http/RequestMatcher/QuerystringRequestMatcher.php
```diff
@@ -45,6 +45,6 @@ public function matches(Request $request): bool
             return true;
         }
 
-        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString()));
+        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString() ?? ''));
```
dbu, line 48: i think the $request must always exist and have the method, so this should be good enough:

```suggestion
        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString() ?: ''));
```

dbu (commented):
thanks for the fixes. one suggestion to be a bit more robust i think.

src/Http/RequestMatcher/QuerystringRequestMatcher.php
cboucaut: Thank, i updated my PR with your suggestion :)

dbu:
thanks a lot!

PR #580, opened 2022-08-03
fix(deprecation) Fix warning about service config

thePanz:
This fixes the warning reported by SF 5.4 when using the legacy context (this is similar to what done in user_context_legacy_sf6.xml)

Humm, looks like SF 4.x complains about an invalid element.. will check that later :)


dbu:
i think you can't do that in symfony 4, which is why the build with symfony 4 now fails. if you want to fix it, you need to add another copy of the file for symfony 5.4 specifically. as its just a deprecation warning, i am not sure its worth the bother but if you want to do it, it needs a 5.4 specific file.

thePanz:
Not worth the time IMO
Will close this PR :)

dbu:
thanks for looking into it. agree, as long as we don't have warnings with symfony 6 i think we are fine here.

PR #581, opened 2022-08-04
Fix several deprecation notices

XWB:
This fixes several deprecation notices:

```
Method "Twig\Extension\ExtensionInterface::getFunctions()" might add "array" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\Twig\CacheTagExtension" now to avoid errors or add an explicit @return annotation to suppress this message.

Method "Symfony\Component\EventDispatcher\EventSubscriberInterface::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\EventListener\CacheControlListener" now to avoid errors or add an explicit @return annotation to suppress this message.

Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.

Method "Symfony\Component\DependencyInjection\Extension\ConfigurationExtensionInterface::getConfiguration()" might add "?ConfigurationInterface" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\DependencyInjection\FOSHttpCacheExtension" now to avoid errors or add an explicit @return annotation to suppress this message. 

User Deprecated: Method "Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationInterface::getAliasName()" might add "string" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\Configuration\Tag" now to avoid errors or add an explicit @return annotation to suppress this message. 

User Deprecated: Method "Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationInterface::allowArray()" might add "bool" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\Configuration\Tag" now to avoid errors or add an explicit @return annotation to suppress this message. 
```

### DIFF
--- a/src/DependencyInjection/FOSHttpCacheExtension.php
+++ b/src/DependencyInjection/FOSHttpCacheExtension.php
@@ -17,6 +17,7 @@ use FOS\HttpCache\SymfonyCache\KernelDispatcher;
 use FOS\HttpCache\TagHeaderFormatter\MaxHeaderValueLengthFormatter;
 use FOS\HttpCacheBundle\DependencyInjection\Compiler\HashGeneratorPass;
 use FOS\HttpCacheBundle\Http\ResponseMatcher\ExpressionResponseMatcher;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Console\Application;
@@ -37,7 +38,7 @@ class FOSHttpCacheExtension extends Extension
     /**
      * {@inheritdoc}
      */
-    public function getConfiguration(array $config, ContainerBuilder $container)
+    public function getConfiguration(array $config, ContainerBuilder $container): ConfigurationInterface
     {
         return new Configuration($container->getParameter('kernel.debug'));
     }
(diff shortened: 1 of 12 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -37,7 +38,7 @@ class FOSHttpCacheExtension extends Extension
     /**
      * {@inheritdoc}
      */
-    public function getConfiguration(array $config, ContainerBuilder $container)
+    public function getConfiguration(array $config, ContainerBuilder $container): ?ConfigurationInterface
```
dbu, line 41:
```suggestion
    public function getConfiguration(array $config, ContainerBuilder $container): ConfigurationInterface
```

i think we are allowed to be more strict about the return type than the parent class. we know we always return an instance and never null

src/Twig/CacheTagExtension.php
```diff
@@ -55,9 +55,6 @@ public function addTag($tag)
         $this->responseTagger->addTags((array) $tag);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getName()
```
dbu, line 58:
```suggestion
    public function getName(): string
```

dbu (commented):
thanks, looks good. i added two suggestions, can you check if you agree with these?

XWB:
@dbu Done :)

dbu:
thank you!

PR #582, opened 2022-08-31
Replace deprecated Command::$defaultName

XWB:
The use of `Command::$defaultName` is deprecate since Symfony 6.1: https://github.com/symfony/symfony/blob/6.2/UPGRADE-6.1.md

### DIFF
--- a/src/Command/InvalidatePathCommand.php
+++ b/src/Command/InvalidatePathCommand.php
@@ -12,6 +12,7 @@
 namespace FOS\HttpCacheBundle\Command;
 
 use FOS\HttpCacheBundle\CacheManager;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -21,11 +22,10 @@ use Symfony\Component\Console\Output\OutputInterface;
  *
  * @author David Buchmann <mail@davidbu.ch>
  */
+#[AsCommand(name: 'fos:httpcache:invalidate:path')]
 class InvalidatePathCommand extends BaseInvalidateCommand
 {
     use PathSanityCheck;
-
-    protected static $defaultName = 'fos:httpcache:invalidate:path';
 
     /**
      * If no cache manager is specified explicitly, fos_http_cache.cache_manager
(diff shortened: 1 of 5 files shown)


dbu:
the upgrade description is to use the AsCommand annotation. should we add that as well (but keep the `setName` so things still work for symfony < 6.1)? afaik the idea of the whole thing is to allow symfony to show the list of available commands without having to instantiate all the command classes?

XWB:
Well, lazy loading is just a preference, not a requirement. And #AsCommand only works with PHP 8, but I can add it if you want.

dbu:
yep, please do so that we do not have a performance degradation for people on PHP 8 & Symfony 6. it will also prepare us to remove the old way when we drop old versions support.

XWB:
@dbu Done.

src/Command/InvalidatePathCommand.php
```diff
@@ -37,7 +37,7 @@ public function __construct(CacheManager $cacheManager = null)
     {
         if (2 <= func_num_args()) {
             @trigger_error('Passing a command name in the constructor is deprecated and will be removed in version 3', E_USER_DEPRECATED);
-            static::$defaultName = func_get_arg(1);
+            $this->setName(func_get_arg(1));
```
dbu, line 42: i think you need to move this to after the parent::__construct now, otherwise we overwrite the custom name in the `configure` method. when setting the static name, the parent sees the new name, now it no longer sees it.

src/Command/InvalidateRegexCommand.php
```diff
@@ -36,7 +36,7 @@ public function __construct(CacheManager $cacheManager = null, $commandName = 'f
     {
         if (2 <= func_num_args()) {
             @trigger_error('Passing a command name in the constructor is deprecated and will be removed in version 3', E_USER_DEPRECATED);
-            static::$defaultName = func_get_arg(1);
+            $this->setName(func_get_arg(1));
```
dbu, line 41: same here, needs to move to after the parent call

src/Command/InvalidateTagCommand.php
```diff
@@ -35,7 +35,7 @@ public function __construct(CacheManager $cacheManager = null, $commandName = 'f
     {
         if (2 <= func_num_args()) {
             @trigger_error('Passing a command name in the constructor is deprecated and will be removed in version 3', E_USER_DEPRECATED);
-            static::$defaultName = func_get_arg(1);
+            $this->setName(func_get_arg(1));
```
dbu, line 40: and here too, move down

src/Command/RefreshPathCommand.php
```diff
@@ -37,7 +37,7 @@ public function __construct(CacheManager $cacheManager = null)
     {
         if (2 <= func_num_args()) {
             @trigger_error('Passing a command name in the constructor is deprecated and will be removed in version 3', E_USER_DEPRECATED);
-            static::$defaultName = func_get_arg(1);
+            $this->setName(func_get_arg(1));
```
dbu, line 42: and finally here as well

src/Command/InvalidatePathCommand.php
XWB: Done

dbu (approved):
thanks a lot!

PR #587, opened 2023-02-20
Add cloudflare support

jean-gui:
Cloudflare Support as a proxy client has recently been added to fos http cache, but the bundle doesn't allow configuring it.

This PR is an attempt at adding this support. I tried to replicate what's being done for Varnish, but with options meaningful to Cloudflare. 

```yaml
# config/packages/fos_http_cache.yaml
fos_http_cache:
    proxy_client:
        cloudflare:
            zone_identifier: '<my-zone-identifier>'
            authentication_token: '<user-authentication-token>'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Changelog
 2.15.0
 ------
 
-* Add support for Cloudflare proxy client.
+* Added Cloudflare proxy client.
 
 2.14.0
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.15.0
+------
+
+* Add support for Cloudflare proxy client.
+
 2.14.0
 ------
 
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -263,6 +263,14 @@ User API token for authentication against Cloudflare APIs, requires ``Zone.Cache
 
 Identifier for the Cloudflare zone you want to purge the cache for.
 
+``http.servers``
+""""""""""""""""
+
+**type**: ``array`` **default**: ``['https://api.cloudflare.com']``
+
+List of Cloudflare API endpoints to use for purging the cache. You can use this to specify a different
+endpoint for testing purposes.
+
 .. _configuration_noop_proxy_client:
 
 noop
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -254,7 +254,7 @@ cloudflare
 
 **type**: ``string``
 
-User API token for authentication against Cloudflare APIs, requires Zone.Cache Purge permissions.
+User API token for authentication against Cloudflare APIs, requires ``Zone.Cache`` Purge permissions.
 
 ``zone_identifier``
 """""""""""""""""
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -248,6 +248,9 @@ cloudflare
             cloudflare:
                 zone_identifier: '<my-zone-identifier>'
                 authentication_token: '<user-authentication-token>'
+                http:
+                    servers:
+                        - 'https://api.cloudflare.com'
 
 ``authentication_token``
 """""""""""""""""""""""
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -12,7 +12,8 @@ The proxy client is also directly available as a service. The default client
 can be autowired with the ``FOS\HttpCache\ProxyClient\ProxyClient`` type
 declaration or the service ``fos_http_cache.default_proxy_client``. Specific
 clients, if configured, are available as ``fos_http_cache.proxy_client.varnish``
-, ``fos_http_cache.proxy_client.nginx`` or ``fos_http_cache.proxy_client.symfony``).
+, ``fos_http_cache.proxy_client.nginx``, ``fos_http_cache.proxy_client.symfony``
+or ``fos_http_cache.proxy_client.cloudflare``).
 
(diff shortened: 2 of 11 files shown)


Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -236,6 +237,32 @@ HTTP method for sending purge requests to the Symfony HttpCache. Make sure to
...
+            cloudflare:
+                zone_identifier: '<my-zone-identifier>'
+                authentication_token: '<user-authentication-token>'
+
+``authentication_token``
+"""""""""""""""""""""""
+
+**type**: ``string``
+
+User API token for authentication against Cloudflare APIs, requires Zone.Cache Purge permissions.
```
dbu:
```suggestion
User API token for authentication against Cloudflare APIs, requires ``Zone.Cache Purge`` permissions.
```

(is this the name of the permission? or is it `Zone.Cache` Purge?

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -455,19 +458,36 @@ private function loadSymfony(ContainerBuilder $container, XmlFileLoader $loader,
         $loader->load('symfony.xml');
     }
 
+    private function loadCloudflare(ContainerBuilder $container, XmlFileLoader $loader, array $config)
+    {
+        $this->createHttpDispatcherDefinition(
+            $container,
+            array_merge(['servers' => ['https://api.cloudflare.com']], $config['http']),
```
dbu: i wonder if we should allow to configure the domain? not sure if there is a use case for it in testing or such, but it feels a bit odd to hardcode it here.

if we do, we should of course provide this domain as default value so it only needs to be set when something else is wanted.

dbu (approved):
thanks a lot for this! looks good - i commented on some details.

and can you please add a note to the changelog (as 2.15.0, i will release when merged)?

have you tried this branch in an actual application to sanity check if it actually works as expected?

Resources/doc/reference/configuration/proxy-client.rst
jean-gui: The latter, updated in dfc680ab9c45ed8d8d9d0eb9b26dfc44aebe0d96

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.15.0
+------
+
+* Add support for Cloudflare proxy client.
```
dbu:
```suggestion
* Added Cloudflare proxy client.
```

src/DependencyInjection/FOSHttpCacheExtension.php
jean-gui: I hesitated to make servers configurable but I guess it doesn't cost much to allow it, and provide a default value. So it's now possible, see 58a3812605bc0ac887bda90daff359f2bd2192f4 and f630ff1dc100fddc1a2177b8bdfbd4cd2deeb100

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -236,6 +237,40 @@ HTTP method for sending purge requests to the Symfony HttpCache. Make sure to
...
+-------
+
+.. code-block:: yaml
+
+    # config/packages/fos_http_cache.yaml
+    fos_http_cache:
+        proxy_client:
+            cloudflare:
+                zone_identifier: '<my-zone-identifier>'
+                authentication_token: '<user-authentication-token>'
```
dbu, line 250:
```suggestion
                authentication_token: '<user-authentication-token>'
                servers: ['https://api.cloudflare.com']
```

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -236,6 +237,40 @@ HTTP method for sending purge requests to the Symfony HttpCache. Make sure to
...
+User API token for authentication against Cloudflare APIs, requires ``Zone.Cache`` Purge permissions.
+
+``zone_identifier``
+"""""""""""""""""
+
+**type**: ``string``
+
+Identifier for the Cloudflare zone you want to purge the cache for.
+
+``http.servers``
```
dbu, line 269: ~~``servers``~~

jean-gui:
> have you tried this branch in an actual application to sanity check if it actually works as expected?

I have tried it with invalidatePath(), and it seemed to work well.

Resources/doc/reference/configuration/proxy-client.rst
jean-gui: I've used the same http.servers configuration key/value as the one for Varnish. This is the result of  ``./bin/console d:conf FOSHttpCacheBundle``:

```yaml
    proxy_client:
        cloudflare:
            authentication_token: '%env(CLOUDFLARE_AUTH_TOKEN)%'
            zone_identifier: '%env(CLOUDFLARE_ZONE_ID)%'
            http:
                servers:
                    - 'https://api.cloudflare.com'
                http_client: null
```

Do you think I should remove it from from the ``http``` key and only allow setting one server?

dbu: ah sorry, no i think using the same structure is good, reduces cognitive overload.

for example, i could configure a client that sends requests to actual cloudflare and some debuggging endpoint i built, to see what is getting invalidated, etc.

Resources/doc/reference/configuration/proxy-client.rst
dbu: sorry, this was correct then, removing my change request.

Resources/doc/reference/configuration/proxy-client.rst
dbu: can you please update the example in the doc to mention the http section?

jean-gui: Done in 08e127b6

jean-gui: Done in 08e127b6

dbu:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.15.0

dbu:
thanks a lot!

PR #591, opened 2023-05-20
Fix symfony 6.3 return type deprecations

acrobat:
> Method "Symfony\Component\DependencyInjection\Extension\ExtensionInterface::load()" might add "void" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\DependencyInjection\FOSHttpCacheExtension" now to avoid errors or add an explicit @return annotation to suppress this message.

(For all compiler passes)
> Method "Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface::process()" might add "void" as a native return type declaration in the future. Do the same in implementation "FOS\HttpCacheBundle\DependencyInjection\Compiler\LoggerPass" now to avoid errors or add an explicit @return annotation to suppress this message.

> User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::build()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\FOSHttpCacheBundle" now to avoid errors or add an explicit @return annotation to suppress this message.

### DIFF
--- a/src/DependencyInjection/Compiler/HashGeneratorPass.php
+++ b/src/DependencyInjection/Compiler/HashGeneratorPass.php
@@ -26,7 +26,7 @@ class HashGeneratorPass implements CompilerPassInterface
     /**
      * {@inheritdoc}
      */
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         if (!$container->has('fos_http_cache.user_context.hash_generator')) {
             return;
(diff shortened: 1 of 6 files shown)


dbu (changes requested):
thanks!

as the compiler passes of this bundle are not meant as extension points, i would prefer to add the actual `: void` return type declaration to the classes directly instead of the annotations. can you change it to that please?

acrobat:
@dbu sure, done!

dbu (approved):
thank you

dbu:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.16.1

PR #592, opened 2023-06-08
Fix symfony 6.3 return type deprecations

XWB:
Fixes the following deprecation errors:

```
Method "Symfony\Component\Console\Command\Command::configure()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\Command\InvalidatePathCommand" now to avoid errors or add an explicit @return annotation to suppress this message. 

Method "Symfony\Component\Console\Command\Command::configure()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\Command\InvalidateRegexCommand" now to avoid errors or add an explicit @return annotation to suppress this message. 

Method "Symfony\Component\Console\Command\Command::configure()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\Command\RefreshPathCommand" now to avoid errors or add an explicit @return annotation to suppress this message. 

Method "Symfony\Component\Console\Command\Command::configure()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\Command\ClearCommand" now to avoid errors or add an explicit @return annotation to suppress this message. 

Method "Symfony\Component\Console\Command\Command::configure()" might add "void" as a native return type declaration in the future. Do the same in child class "FOS\HttpCacheBundle\Command\InvalidateTagCommand" now to avoid errors or add an explicit @return annotation to suppress this message. 
```

### DIFF
--- a/src/Command/ClearCommand.php
+++ b/src/Command/ClearCommand.php
@@ -29,7 +29,7 @@ class ClearCommand extends BaseInvalidateCommand
     /**
      * {@inheritdoc}
      */
-    protected function configure()
+    protected function configure(): void
     {
         $this
             ->setName('fos:httpcache:clear')
--- a/src/Command/InvalidatePathCommand.php
+++ b/src/Command/InvalidatePathCommand.php
@@ -46,7 +46,7 @@ class InvalidatePathCommand extends BaseInvalidateCommand
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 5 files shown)


dbu:
thanks a lot!

PR #595, opened 2023-08-03
SF6: catch SessionNotFoundException

ro0NL:
Since SF6 `getSession` is throwing, rather than a nullable (but i kept it for BC)

This handles it.

We noticed `Session not found` error on http://, since this listener runs before redirecting to https://


dbu (approved):
good catch, thanks a lot! and yes until we drop support for older versions, we need to still check for null.

dbu:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.16.2

ro0NL:
fastest release ever :rocket: :100:

PR #599, opened 2023-12-13
[DRAFT]: Remove ExtraFrameworkBundle

toxicity1985:
Hello,

I have working on the way to remove the dependencies to ExtraFrameworkBundle.
I create an AnnotationListener to get the annotation from the controller and get a few stuff from ExtraFrameworkBundle.
I have some issue with the test but i have no idea why.
I know that's related to the AbstractRuleListener. The matchRule is not working as expected.
If you have some idea what am i wrong ?
It's a draft version.

image


tests/Unit/Http/RuleMatcherTest.php
```diff
@@ -15,14 +15,15 @@
...
+use Symfony\Component\HttpFoundation\RequestMatcher\AttributesRequestMatcher;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
 use Symfony\Component\HttpFoundation\Response;
 
 class RuleMatcherTest extends TestCase
 {
     public function testRequestMatcherCalled()
     {
-        $requestMatcher = new RequestMatcher(null, null, null, null, ['_controller' => '^AcmeBundle:Default:index$']);
+        $requestMatcher = new AttributesRequestMatcher(['_controller' => '^AcmeBundle:Default:index$']);
```
dbu, line 26: ah right, we need to fix those for symfony 7.

the new class is only available since Symfony 6.2 can you please do `class_exists(AttributesRequestMatcher::class) ? new AttributesRequestMatcher(...) : new RequestMatcher(...);` to keep legacy tests working?

and i would prefer this in a separate pull request as it is not related to extra framework bundle. it is fixing a deprecation in symfony 6 and avoid the error in symfony 7.

dbu:
oh snap. i thought we could get rid of the explicit references to FrameworkExtraBundle but keep using it for the annotations if it is available. but thats because i recently upgraded to attributes in a library that used the doctrine annotations implementation.

i think porting the annotation things from extra bundle into this bundle is adding too much unrelated legacy code. i think after all we should do a new major version that simply drops annotations support. i created the 3.x branch for a new major version.

sorry for the work i made you do - i hope it was at least interesting to see how it works :see_no_evil: 

can you target 3.x with the fixes for the specific request matchers? we don't need the class_exists after all, we should simply make the 3.x version only allow symfony 6.4 and 7. and attributes should be simply deleting the annotation things and they will only have the `#[\Attribute...` information.

toxicity1985:
No worries about the double work. It was ok to go deeper inside this bundle.
So il will remove all the stuff related to annotation and make a new pull request for that.
I will also make a new one for the request matcher issue.

dbu:
thank you very much @toxicity1985 !

toxicity1985:
i can close this one.

PR #600, opened 2023-12-13
remove annotation for 3.X release

toxicity1985:


.github/workflows/php.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           - php-version: '7.4'
           - php-version: '8.0'
```
dbu: symfony 6.4 needs php 8.1 or newer, we should remove the older builds

composer.json
```diff
@@ -21,11 +21,11 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^8.2",
```
dbu: lets do ^8.1 we don't need the very latest and symfony 6.1 allows 8.1

composer.json
```diff
@@ -35,20 +35,19 @@
...
         "mockery/mockery": "^1.3.2",
         "monolog/monolog": "*",
-        "sensio/framework-extra-bundle": "^4.0 || ^5.5.1 || ^6.0",
         "doctrine/annotations": "^1.11",
-        "symfony/browser-kit": "^4.4 || ^5.0 || ^6.0",
-        "symfony/console": "^4.4 || ^5.0 || ^6.0",
-        "symfony/finder": "^4.4 || ^5.0 || ^6.0",
+        "symfony/browser-kit": "^6.0 || ^7.0",
+        "symfony/console": "^6.0 || ^7.0",
+        "symfony/finder": "^6.0 || ^7.0",
```
dbu: lets make the dev symfony dependencies 6.4 as well, not 6.0

src/Configuration/ConfigurationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface ConfigurationInterface
+{
+    /**
+     * Returns the alias name for an annotated configuration.
+     *
+     * @return string
+     */
+    public function getAliasName();
```
dbu:
```suggestion
    public function getAliasName(): string;
```

we now have modern php, so we can use return type declarations. (and remove the redundant `@return string` annotation)

src/Http/ResponseMatcher/ExpressionResponseMatcher.php
```diff
@@ -11,7 +11,7 @@
 
 namespace FOS\HttpCacheBundle\Http\ResponseMatcher;
 
-use Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage as SecurityExpressionLanguage;
+use FOS\HttpCacheBundle\Security\ExpressionLanguage as SecurityExpressionLanguage;
```
dbu: do we still need this class or can we remove it in favor of https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpFoundation/RequestMatcher/ExpressionRequestMatcher.php

tests/Functional/EventListener/InvalidationListenerTest.php
```diff
@@ -19,83 +19,6 @@ class InvalidationListenerTest extends WebTestCase
 {
     use MockeryPHPUnitIntegration;
 
-    public function testInvalidateRoute()
```
dbu: do we have corresponding tests with attributes? if not we should convert this to test attributes so we have functional tests of the thing working in action.

dbu (commented):
thanks a lot!

i added some comments.

.github/workflows/php.yml
dbu: but lets keep lowest builds, just with the lowest supported php version

composer.json
toxicity1985: yeah but i switch to symfony 7 and symfony 7 needs php8.2 no ?

dbu: we allow symfony 6.4 || 7, so this is fine.

composer will make sure that symfony 7 is not installed when on php 8.1, but its maybe helpful for people who for the moment are still on php 8.1 / symfony 6.4.

toxicity1985: Ok. I didn't know that. Thanks.

tests/Functional/EventListener/InvalidationListenerTest.php
toxicity1985: Yes. The test alreay exist on attribute.

toxicity1985:
There is 2 bundle to be updated to be comptabile with php8.2.

toxicity1985:
Do you have an idea why the rule matcher is not ok ?
Something has change but i don't found what.

composer.json
```diff
@@ -21,11 +21,11 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
-        "friendsofsymfony/http-cache": "^2.15",
-        "symfony/framework-bundle": "^4.4.0 || ^5.0 || ^6.0",
-        "symfony/http-foundation": "^4.4.0 || ^5.0 || ^6.0",
-        "symfony/http-kernel": "^4.4.0 || ^5.0 || ^6.0"
+        "php": ">= 8.1",
```
dbu:
```suggestion
        "php": "^8.1",
```

i know symfony itself does this, but i strongly disagree. major PHP version changes can do BC breaks, so we should not claim compatibility with all future versions of PHP.

composer.json
```diff
@@ -35,21 +35,20 @@
         "guzzlehttp/guzzle": "^7.2",
         "mockery/mockery": "^1.3.2",
         "monolog/monolog": "*",
-        "sensio/framework-extra-bundle": "^4.0 || ^5.5.1 || ^6.0",
         "doctrine/annotations": "^1.11",
```
dbu: do we still need this?

src/Http/RequestMatcher/QuerystringRequestMatcher.php
```diff
@@ -12,18 +12,23 @@
...
  */
-class QuerystringRequestMatcher extends SymfonyRequestMatcher
+class QuerystringRequestMatcher implements RequestMatcherInterface
 {
     /**
      * @var string Regular expression to match the query string part of the request url
      */
     private $queryString;
 
+    public function __construct($queryString = null)
```
dbu: when we add this to the constructor, i'd like to remove the setter for queryString.

composer.json
```diff
@@ -35,21 +35,20 @@
...
-        "symfony/security-bundle": "^4.4 || ^5.0 || ^6.0",
-        "symfony/twig-bundle": "^4.4 || ^5.0 || ^6.0",
-        "twig/twig": "^2.13",
-        "symfony/yaml": "^4.4 || ^5.0 || ^6.0",
-        "symfony/css-selector": "^4.4 || ^5.0 || ^6.0",
-        "symfony/expression-language": "^4.4 || ^5.0 || ^6.0",
+        "symfony/browser-kit": "^6.4 || ^7.0",
+        "symfony/console": "^6.4 || ^7.0",
+        "symfony/finder": "^6.4 || ^7.0",
+        "symfony/phpunit-bridge": "^7.0",
```
dbu: we could try forcing the latest patch version of the bridge, maybe that helps:

```suggestion
        "symfony/phpunit-bridge": "^7.0.1",
```

src/Configuration/ConfigurationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * ConfigurationInterface.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface ConfigurationInterface
+{
+    /**
+     * Returns the alias name for an annotated configuration.
```
dbu: we should talk about attribute, not annotation.

src/Configuration/ConfigurationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+namespace FOS\HttpCacheBundle\Configuration;
+
+/**
+ * ConfigurationInterface.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface ConfigurationInterface
```
dbu: looking at how we use this in Php8AttributesListener i would prefer to do a `match` on $instance to see if it is one of the 3 known attribute classes and if so add it with the corresponding alias name. this interface is not really adding value anymore i think.

src/Command/BaseInvalidateCommand.php
```diff
@@ -24,8 +23,6 @@
  */
 abstract class BaseInvalidateCommand extends Command
 {
-    use ContainerAwareTrait;
```
dbu: do we need to add a setContainer method and call setContainer in our services?

the recommended way in the upgrade instructions is to not rely on the container anymore - would that be possible here, can we figure out what our code does with the container and inject the specific things to the specific commands?

(that would be out of scope for this MR though. if commands still work with fetching the container from the application object we can merge like this for now)

dbu (commented):
thanks for the work!

i try to [get the ball rolling](https://github.com/Jean-Beru/fos-http-cache-cloudfront/pull/3) with the cloudfront extension. worst case we disable the test if the component is not updated for the new version soonish.

tests/Functional/EventListener/InvalidationListenerTest.php
dbu: great!

src/Command/BaseInvalidateCommand.php
dbu: i created #604 for this

toxicity1985:
Hi,

I update the plugin this morning. I add a match function as asked but the get alias is necessary to add configuration inside the request. I juste have a question regards the invalidation path attribute. If we have a match we have to clear the path configured inside the attribute but not the route called ?

PR #605, opened 2023-12-18
Fix compiling container using Secrets

DemigodCode:
If one wants to use secrets as parameters for fastly or cloudflare, the compiling of the container will fail.
After compiling the container the termination event will cause this error cause it's not necessary to have the decrypt key available when compiling the container and the service can't be initialized.

This fix creates the services as lazy services which will not make the options necessary on compile time of the container.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "friendsofsymfony/http-cache": "^2.15",
         "symfony/framework-bundle": "^4.4.0 || ^5.0 || ^6.0",
         "symfony/http-foundation": "^4.4.0 || ^5.0 || ^6.0",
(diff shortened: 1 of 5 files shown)


DemigodCode:
Don't know why readthedocs is failing, did not change anything there.

Also don't know why the unit tests are failing cause of simple-phpunit. That's maybe something else?

dbu:
thanks for this pull request. indeed, makes sense to make the services lazy to avoid the problem you describe.

we are close to release version 3 of the bundle for new symfony versions. the build seems to have rotten again on the 2.x branch. could you base this PR onto the 3.x branch? then it will be in the upcoming release and the CI should work out.

PR #606, opened 2023-12-19
Fix compiling container using Secrets

DemigodCode:
If one wants to use secrets as parameters for fastly or cloudflare, the compiling of the container will fail.
After compiling the container the termination event will cause this error cause it's not necessary to have the decrypt key available when compiling the container and the service can't be initialized.

This fix creates the services as lazy services which will not make the options necessary on compile time of the container.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "friendsofsymfony/http-cache": "^2.15",
         "symfony/framework-bundle": "^4.4.0 || ^5.0 || ^6.0",
         "symfony/http-foundation": "^4.4.0 || ^5.0 || ^6.0",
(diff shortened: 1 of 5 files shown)


DemigodCode:
@dbu same in 3.x closed 2.x PR, don't know how to rebase it and created this one.

dbu:
oh sorry. i mixed up FOSHttpCache where i fixed things and FOSHttpCacheBundle :see_no_evil: 

i fixed the build in #608 and then merged things up to 3.x - can you please rebase on the 3.x branch?.

DemigodCode:
Done

dbu:
thanks a lot! merged in #609

PR #610, opened 2023-12-20
remove annotation for 3.X release

toxicity1985:
Hello,

Sorry about the mess but i make something wrong with the rebase and drop the commit.
I update the plugin yesterday. I add a match function as asked.
I juste have a question regards the invalidation path attribute. If we have a match we have to clear the path configured inside the attribute but not the route called ?


dbu:
> If we have a match we have to clear the path configured inside the attribute but not the route called ?

yes. the assumption is that you annotate this on a route that changes content (so usually it will be a POST request). if the request that triggers the invalidation would itself be cached, symfony would not be invoked on a cache hit and the invalidation would not be triggered.

composer.json
```diff
@@ -35,22 +35,21 @@
         "guzzlehttp/guzzle": "^7.2",
         "mockery/mockery": "^1.3.2",
         "monolog/monolog": "*",
-        "sensio/framework-extra-bundle": "^4.0 || ^5.5.1 || ^6.0",
         "doctrine/annotations": "^1.11",
-        "symfony/browser-kit": "^4.4 || ^5.0 || ^6.0",
-        "symfony/console": "^4.4 || ^5.0 || ^6.0",
-        "symfony/finder": "^4.4 || ^5.0 || ^6.0",
-        "phpunit/phpunit": "^9.6.15",
```
dbu, line 43: please don't revert this change. i ditched phpunit-bridge in favor of directly using phpunit.

toxicity1985: i will check. i just accept my modification when i rebase the branch.

composer.json
```diff
@@ -35,20 +35,19 @@
         "guzzlehttp/guzzle": "^7.2",
         "mockery/mockery": "^1.3.2",
         "monolog/monolog": "*",
-        "sensio/framework-extra-bundle": "^4.0 || ^5.5.1 || ^6.0",
         "doctrine/annotations": "^1.11",
```
dbu, line 38: i think we don't need this anymore

dbu:
thanks. looks like the CI is now reporting properly again. ignore the installation problem with jean-beru for now.

the failing tests in the other builds indicate that something is not yet working properly with the attributes. or the fixtures only have annotations not attributes?

dbu:
@toxicity1985 do you have time to check why the build fails?

toxicity1985:
Hello,

I will check today or tommorow. I was busy last week :-p

dbu:
thanks - did not want to rush you, just to check if it went under or if you are still aware of it.

toxicity1985:
I fixed some test but the last are related to the cache maybe something related to configuration but i don't find where.

toxicity1985:
@dbu i fixed the last test.

usu:
Hi @dbu & @toxicity1985. Is there anything than could be supported to test or to fix any of the remaining tests? Or is this ready to be merged?

toxicity1985:
Hello @usu,
I think it's ok. I use the fork version on my project.
But i think we also have to update some parent library.

composer.json
```diff
@@ -21,11 +21,11 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
-        "friendsofsymfony/http-cache": "^2.15",
-        "symfony/framework-bundle": "^4.4.0 || ^5.0 || ^6.0",
-        "symfony/http-foundation": "^4.4.0 || ^5.0 || ^6.0",
-        "symfony/http-kernel": "^4.4.0 || ^5.0 || ^6.0"
+        "php": "^8.1",
+        "friendsofsymfony/http-cache": "^3.x-dev",
```
dbu: can we do `^2.15||3.x-dev` instead? the jeanberu integration does not allow the (not yet released http-cache 3)

composer.json
```diff
@@ -35,20 +35,19 @@
         "guzzlehttp/guzzle": "^7.2",
         "mockery/mockery": "^1.3.2",
```
dbu: the lowest build fails with a fatal error on prophecy. is there a newer version of mockery that we can require as minimal version to avoid the issue? (i think this is where prophecy is coming from)

dbu (commented):
sorry for the silence. i hope i can release 3.x soon. 

there is an open comment still, and some issues with the ci.

(codestyle and readthedocs is out of scope, we can ignore those. but the phpunit builds should all be green)

toxicity1985:
I have to fix the style. But all the test pass now.

src/Security/ExpressionLanguage.php
```diff
@@ -0,0 +1,33 @@
...
+namespace FOS\HttpCacheBundle\Security;
+
+use Symfony\Component\Security\Core\Authorization\ExpressionLanguage as BaseExpressionLanguage;
+
+/**
+ * Adds some function to the default Symfony Security ExpressionLanguage.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class ExpressionLanguage extends BaseExpressionLanguage
```
dbu, line 21: is this replacing something else?

composer.json
dbu: thanks

src/Security/ExpressionLanguage.php
toxicity1985: Yes,, something used in the framework bundle.

dbu:
thanks a lot! i wrapped this up in #611

there are still a couple of things left in the version 3 milestone https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/milestone/6

PR #611, opened 2024-03-20
remove annotation for 3.x release

dbu:
continuation of #610


src/Security/ExpressionLanguage.php
```diff
@@ -0,0 +1,33 @@
...
+namespace FOS\HttpCacheBundle\Security;
+
+use Symfony\Component\Security\Core\Authorization\ExpressionLanguage as BaseExpressionLanguage;
+
+/**
+ * Adds some function to the default Symfony Security ExpressionLanguage.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class ExpressionLanguage extends BaseExpressionLanguage
```
dbu: see discussion at https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/610#discussion_r1531560480

i don't think this should be in the FOSHttpCacheBundle, it has nothing directly to do with it.

and reading https://symfony.com/doc/current/security/expressions.html i would expect those expressions to be supported by core symfony. can we remove it?

src/Http/RequestMatcher/QuerystringRequestMatcher.php
```diff
@@ -45,6 +41,14 @@ public function matches(Request $request): bool
             return true;
         }
 
-        return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString() ?: ''));
+        if ($request->getQueryString()) {
+            return (bool) preg_match('{'.$this->queryString.'}', rawurldecode($request->getQueryString() ?: ''));
+        }
+
+        if ($request->getRequestUri()) {
+            return (bool) preg_match('#'.$this->queryString.'#', $request->getRequestUri());
```
dbu: this makes the query string matcher match on request uri when there is no query string. i don't think that is correct

dbu: also, with the check if query string is null, we should never end up here.

src/Security/ExpressionLanguage.php
dbu: asked in symfony slack if there is any service we can use to get the "full" expression language of the application

dbu: went with allowing a custom service and defaulting to the symfony/expression-language default implementation.

PR #618, opened 2024-03-28
Remove deprecated code

dbu:


usu:
I found one more reference to the deprecated `fos_http_cache.user_context.logout_handler` in [ServiceTest](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/tests/Functional/DependencyInjection/ServiceTest.php#L56)

usu:
~~Found some more legacy code in Configuration.php:~~
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/src/DependencyInjection/Configuration.php#L59
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/src/DependencyInjection/Configuration.php#L620
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/src/DependencyInjection/Configuration.php#L657

~~Also, there are quite some errors with phpstan in this file. Didn't look into it further other than typing errors only.~~

Edit: included in #621

usu:
~~And finally in FlashMessageListener.php, I think the `class_alias` could just be removed and the reference to the alias further below replaced with the actual `ResponseEvent`.~~

https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/src/EventListener/FlashMessageListener.php#L22
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/3.x/src/EventListener/FlashMessageListener.php#L57

Edit: included in #622

dbu:
hooray, the phpunit failure we previously had here went away with the rebase on 3.x

PR #620, opened 2024-04-08
Allow usage of friendsofsymfony/http-cache ^3.0

alexander-schranz:
I think composer does not interpret `3.x-dev` as `3.0-dev`.

Bildschirmfoto 2024-04-08 um 14 36 38

I recommend to use `composer config minimum-stability dev` ([has no effects to projects](https://getcomposer.org/doc/04-schema.md#minimum-stability)) or we go with `^2.15 || ^3.0 || ^3.0@dev`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "friendsofsymfony/http-cache": "^2.15 || 3.x-dev",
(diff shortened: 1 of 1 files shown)


alexander-schranz:
Error on my side it seems also work with `3.x-dev`

dbu:
:sweat_smile: yeah i think there was a different reason why 3 was not picket up.

PR #621, opened 2024-05-01
fix: static typing and final classes

### LINKED ISSUE #603: static typing
convert all code to use static typing, make internal classes final.

usu:
Fixes #603

### Mark internal classes final
I sticked to the rules outlined by Ocramius in https://ocramius.github.io/blog/when-to-declare-classes-final/

### Static typing
Fixes most typing related errors from `phpstan analyze -l 6 src'
- excluding typing of items within arrays (https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type)
- excluding non-typing related errors

### DIFF
--- a/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
+++ b/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
@@ -24,7 +24,7 @@ use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
  * This handler is deprecated because it never did what it was supposed to do. The session is already invalidated by the SessionLogoutHandler
  * which is always the first logout handler executed
  */
-class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
+final class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
 {
     private $invalidator;
 
--- a/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
+++ b/src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
@@ -24,7 +24,7 @@ use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
  * This handler is deprecated because it never did what it was supposed to do. The session is already invalidated by the SessionLogoutHandler
  * which is always the first logout handler executed
  */
-final class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
+class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
 {
     private $invalidator;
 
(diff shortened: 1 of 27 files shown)


usu:
I also included removal of legacy code in Configuration.php as it was a mix of legacy code and static typing.

With extension phpstan/phpstan-symfony, this now passes Level 6 with one remaining error only (which is obvious, as the CloudFront package is only included as a `suggest`, not as dependency).

```
 ------ ---------------------------------------------------------------------
  Line   Configuration.php
 ------ ---------------------------------------------------------------------
  593    Class JeanBeru\HttpCacheCloudFront\Proxy\CloudFront not found.
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
 ------ ---------------------------------------------------------------------
```

src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
```diff
@@ -24,7 +24,7 @@
  * This handler is deprecated because it never did what it was supposed to do. The session is already invalidated by the SessionLogoutHandler
  * which is always the first logout handler executed
  */
-final class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
+class ContextInvalidationLogoutHandler implements LogoutHandlerInterface
```
dbu: why do we need to un-final this one?

dbu (approved):
thanks a lot for this one! just one small question.

we should add a ci step for phpstan similar to FOSHttpCache https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/.github/workflows/static.yml

(where we could require the optional dependency, or alternatively tell phpstan about its possible existence)

usu:
> we should add a ci step for phpstan similar to FOSHttpCache https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/.github/workflows/static.yml
> 
> (where we could require the optional dependency, or alternatively tell phpstan about its possible existence)

Yeah, fully agree. I can make a seperate PR for this.

src/Security/Http/Logout/ContextInvalidationLogoutHandler.php
usu: Oh, this was definitely not intended. Thanks for noticing. Just reverted with commit https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/621/commits/2a8f76d0e5bfabd878f3b102d2e1ac440761df52

This class will then be removed with PR https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/618

dbu:
thanks a lot!

PR #622, opened 2024-05-02
chore: add phpstan to CI

usu:
- Adds phpstan to CI (didn't test it - hope it goes through 😄). Yaml is copied from FOSHttpCache
- Fixes all level-1 errors, except one remaining error which will be fixed by #618
- After that, we can fix and increase higher levels by separate PRs

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,39 @@
+name: Static analysis
+
+on:
+  push:
+    branches:
+      - "*.x"
+  pull_request:
+
+jobs:
+  phpstan-src:
+    name: PHPStan src
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Pull in optional dependencies
+        run: composer require --no-update jean-beru/fos-http-cache-cloudfront
+
+      - name: PHPStan
+        uses: docker://oskarstark/phpstan-ga
+        with:
+          args: analyze --no-progress
+
+  phpstan-tests:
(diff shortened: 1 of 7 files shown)


usu:
The new workflow "static" triggers quite a lot of errors "unknown class", which is due to references to dependencies which are dev-dependencies.
I could easily fix this by adding a `REQUIRE_DEV: "true"` to the `phpstan-src` job. But I wonder if this is correct, or whether all these dependencies should actually be non-dev dependencies...

dbu:
thanks for looking into this.

dang. good old "optional dependencies" saying hello :-| 

i am not sure how optional they really are, a couple places looks to me like it would simply error if thats not available. lets move console and security core & http to the requirements. i would assume most symfony applications use those anyways. and require twig like you require jean-beru - twig really is an optional integration.

usu:
Fixed as suggested & ready to merge. 

Actually console is optional, the services are only loaded when `Symfony\Component\Console\Application` is available:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/2.x/src/DependencyInjection/FOSHttpCacheExtension.php#L105
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/2.x/src/DependencyInjection/FOSHttpCacheExtension.php#L541

However, twig is currently not treated optional. I will fix this in a separate PR.

dbu:
the remaining failure is for ContextInvalidationLogoutHandler which was symfony 5 only and is being deleted in #618 

i merge this as-is with the phpstan failure. thanks for the help!

PR #624, opened 2024-05-03
fix: treat Twig as optional

usu:
Only load CacheTagExtension service in case Twig is available


dbu (approved):
great, thanks!

i will add a note in the changelog after merging

Issue #634, opened 2024-11-22
Docs refer to "latest" but should refer to "stable" instead


mvhirsch:
By clicking on the link on [github's page](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle)  under "about"[foshttpcachebundle.readthedocs.io](http://foshttpcachebundle.readthedocs.io/) I'll get redirected to "latest" documentation.

![image](https://github.com/user-attachments/assets/7c2f1a17-a4b3-439d-8901-2179e2c384c0)

 But "latest" [suggests to install an abandoned bundle `sensio/framework-extra-bundle`](https://foshttpcachebundle.readthedocs.io/en/latest/overview.html#requirements).

The proper documentation for v3 will be found under ["stable"](https://foshttpcachebundle.readthedocs.io/en/stable/overview.html). Is it possible to reconfigure the redirection, so users see "stable" instead of "latest" first? Or is it possible to point "latest" on [default branch `3.x`](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/branches)?

dbu:
i changed the configuration on readthedocs to show `stable` by default. thanks for updating the config, i hope latest now shows the right thing too.

mvhirsch:
@dbu I now get redirected to _stable_ which solves this issue. Thank you very much! 

Unfortunately [latest still suggests](https://foshttpcachebundle.readthedocs.io/en/latest/overview.html#requirements) to install SensioFrameworkExtraBundle. Looks like #635 did not fix it. Any ideas?

PR #635, opened 2024-11-22
Point documentation at `3.x`

mvhirsch:
Not sure if it works. Refs #634


dbu (approved):
i am not sure what the effect is, if this gets visible anywere. but the test build on readthedoc looks fine, so lets merge, it certainly seems correct.

PR #638, opened 2025-01-08
feat: allow to configure header name for reverse_proxy_ttl specific value

94noni:
from https://github.com/FriendsOfSymfony/FOSHttpCache/pull/591

Fastly allows to cache via this https://www.fastly.com/documentation/reference/glossary/#term-surrogate-control header `Surrogate-Control`
(see also https://docs.fastly.com/en/guides/controlling-caching#setting-different-ttls-for-fastly-cache-and-web-browsers)

a specific origin response header, and this bundle allows to define a default value

this PR is allows to configure the ttl header name in the bundle avoid to create a fastly vcl or compute code for example

PR #639, opened 2025-01-09
Fix php error

richardhj:
Fixes #638

### DIFF
--- a/src/DependencyInjection/FOSHttpCacheExtension.php
+++ b/src/DependencyInjection/FOSHttpCacheExtension.php
@@ -17,6 +17,7 @@ use FOS\HttpCache\ProxyClient\ProxyClient;
 use FOS\HttpCache\SymfonyCache\KernelDispatcher;
 use FOS\HttpCache\TagHeaderFormatter\MaxHeaderValueLengthFormatter;
 use FOS\HttpCacheBundle\DependencyInjection\Compiler\HashGeneratorPass;
+use FOS\HttpCacheBundle\EventListener\CacheControlListener;
 use FOS\HttpCacheBundle\Http\ResponseMatcher\ExpressionResponseMatcher;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
@@ -48,7 +49,7 @@ final class FOSHttpCacheExtension extends Extension
 
         if ($config['debug']['enabled'] || (!empty($config['cache_control']))) {
             $debugHeader = $config['debug']['enabled'] ? $config['debug']['header'] : false;
-            $ttlHeader = $config['cache_control']['ttl_header'] ?? null;
+            $ttlHeader = $config['cache_control']['ttl_header'] ?? CacheControlListener::DEFAULT_TTL_HEADER_NAME;
             $container->setParameter('fos_http_cache.debug_header', $debugHeader);
             $container->setParameter('fos_http_cache.ttl_header', $ttlHeader);
             $loader->load('cache_control_listener.xml');
--- a/src/DependencyInjection/FOSHttpCacheExtension.php
+++ b/src/DependencyInjection/FOSHttpCacheExtension.php
@@ -48,7 +48,7 @@ final class FOSHttpCacheExtension extends Extension
 
         if ($config['debug']['enabled'] || (!empty($config['cache_control']))) {
             $debugHeader = $config['debug']['enabled'] ? $config['debug']['header'] : false;
-            $ttlHeader = $config['cache_control']['ttl_header'];
+            $ttlHeader = $config['cache_control']['ttl_header'] ?? null;
             $container->setParameter('fos_http_cache.debug_header', $debugHeader);
             $container->setParameter('fos_http_cache.ttl_header', $ttlHeader);
             $loader->load('cache_control_listener.xml');
--- a/src/EventListener/CacheControlListener.php
+++ b/src/EventListener/CacheControlListener.php
@@ -28,6 +28,8 @@ use Symfony\Component\HttpKernel\KernelEvents;
  */
 final class CacheControlListener implements EventSubscriberInterface
 {
+    public const DEFAULT_TTL_HEADER_NAME = 'X-Reverse-Proxy-TTL';
+
     /**
      * Whether to skip this response and not set any cache headers.
      */
@@ -56,7 +58,7 @@ final class CacheControlListener implements EventSubscriberInterface
          * @var string|false Name of the header or false to add no header
          */
         private readonly string|false $debugHeader = false,
-        private readonly string $ttlHeader = 'X-Reverse-Proxy-TTL',
+        private readonly string $ttlHeader = self::DEFAULT_TTL_HEADER_NAME,
     ) {
     }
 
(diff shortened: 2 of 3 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -48,7 +48,7 @@ public function load(array $configs, ContainerBuilder $container): void
 
         if ($config['debug']['enabled'] || (!empty($config['cache_control']))) {
             $debugHeader = $config['debug']['enabled'] ? $config['debug']['header'] : false;
-            $ttlHeader = $config['cache_control']['ttl_header'];
+            $ttlHeader = $config['cache_control']['ttl_header'] ?? null;
```
94noni: arf I was unsure so I wanted to keep `''` but at the end we removed it 
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/638#discussion_r1907479582

if null is ok here, and the configuration will swap with default value if undefined, its perfect then 👍🏻

94noni (approved):

alexander-schranz:
~~Think the `CacheControlListener` need also be adopted. I'm getting:~~

> TypeError: FOS\HttpCacheBundle\EventListener\CacheControlListener::__construct(): Argument https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/2 ($ttlHeader) must be of type string, null given, called in /home/runner/work/sulu/sulu/src/Sulu/Bundle/WebsiteBundle/Tests/Application/var/cache/website/test/ContainerY4qBT6C/Sulu_Bundle_WebsiteBundle_Tests_Application_KernelTestDebugContainer.php on line 530

Update see: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/639#discussion_r1908347191

src/DependencyInjection/FOSHttpCacheExtension.php
alexander-schranz: This is strange it should be always `X-Reverse-Proxy-TTL` as in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/7b737f770941bfb67e065d1be5975aa26a3ee2d7/src/DependencyInjection/Configuration.php#L280-L283 defined.

Looks for me there is an issue in applying the correct defaults from the config. /cc @dbu

We could go with `$config['cache_control']['ttl_header'] ?? 'X-Reverse-Proxy-TTL';` but that would just a workaround we should find out why the default value don't work.

alexander-schranz: Looks for me in:

```
    ->children()
                ->arrayNode('cache_control')
```

The `->addDefaultsIfNotSet()` is missing.

But this makes the `(!empty($config['cache_control']))` check unnecessary or need to be changed to something else.

:warning: In case of Sulu `fos_http_cache.event_listener.cache_control` was never registered under `prod` environment. So to keep things as they were I think the check (`if ($config['debug']['enabled'] || (!empty($config['cache_control']))) {`) need to be changed, or:

To keep things as they are we could remove `->defaultValue('X-Reverse-Proxy-TTL') ` and go here with `$ttlHeader = $config['cache_control']['ttl_header'] ?? 'X-Reverse-Proxy-TTL';`

94noni: `$config['cache_control']['ttl_header'] ?? 'X-Reverse-Proxy-TTL';` was even my first idea, after we went for nothing on the original PR

yes perhaps in the meantime we can just fallback in the extension with same value as configuration

richardhj: IMHO `addDefaultsIfNotSet()` should be the correct way 🤔

94noni: but its not an array node as per the doc

94noni: to me the better "hot fix" would be to do like [initially done](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/638#discussion_r1907315992)

wdyt?

alexander-schranz: but that would register always the service which is currently not configured `cache_control` for existing projects without cache_contorl settings. So for BC reasons we should not go with `addDefaultsIfNotSet` or we need to change the `if` check.

alexander-schranz: Example we could go if we want the `addDefaultsIfNotSet` the `!empty` would need to be changed to:

```php
        if ($config['debug']['enabled']
           || count($config['cache_control']) > 1 // the ttl_header is always defined with a defaultValue so we need > 1
        ) {
            $debugHeader = $config['debug']['enabled'] ? $config['debug']['header'] : false;
            $ttlHeader = $config['cache_control']['ttl_header'];
```

dbu:
damn, did not realize how defaults in the configuration work. and looks like we did not add enough tests to discover this issue, sorry.

to avoid side effects like always enabling the cache control listener even when not wanted, lets use the fallback in the FOSHttpCacheExtension class as @94noni originally proposed with the default value if nothing is in the array.

the default will then only matter if people write `ttl_header: ~` but afaik it is best practice to do that.

to make things more explicit, lets add a public constant in the src/EventListener/CacheControlListener.php class and reference that both from Configuration and from FOSHttpCacheExtension to avoid the magic string. do you want to do that @richardhj ? then i can tag a patch version with the fix.

src/EventListener/CacheControlListener.php
```diff
@@ -56,7 +58,7 @@ public function __construct(
          * @var string|false Name of the header or false to add no header
          */
         private readonly string|false $debugHeader = false,
-        private readonly string $ttlHeader = 'X-Reverse-Proxy-TTL',
```
richardhj, line 59: This is kinda unnecessary because always set explicitly, but does not hurt.

richardhj:
Should be ready to review

src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -48,7 +49,7 @@ public function load(array $configs, ContainerBuilder $container): void
 
         if ($config['debug']['enabled'] || (!empty($config['cache_control']))) {
```
alexander-schranz, line 50: @dbu Do we want to have the `CacheControlListener` registered if only the new `ttl_header` config is defined, for maybe just the debug ENVs?

If not we would need to go with: `count($config['cache_control']) > 1` and `addDefaultsIfNotSet` in the config. The `??` would not be required then also.

94noni (approved):
lets do it that way in fine 👍🏻

src/EventListener/CacheControlListener.php
dbu: it is relevant for BC in case somebody configured their own service with the class.

src/DependencyInjection/FOSHttpCacheExtension.php
dbu: i can't see a use case for configuring the ttl_header but not configuring any reverse_proxy_ttl at all.

if we merge as is now, my understanding is that the behaviour is the same as before, except that the ttl_header configuration is respected if overwritten, but does not cause errors if not explicitly mentioned by the user. am i missing something?

alexander-schranz: yes it will not break, it just register a listener which does then nothing as no rules are configured.

dbu: i clarified with alexander, we leave it like this. the cleanest way would be to count if there are any cache_control.rules, but this is not the bug and could have edge case side effects if somebody relies on the cache control service to exist even without rules if there is any configuration.

could be a cleanup for 4.x. when we count the rules, we can also addDefaultsIfNotSet on the ttl_header

dbu:
tagged https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/3.1.1

PR #640, opened 2025-01-30
[RFC] feat: introduce new reverse_proxy_cache_control config to allow configuring specific cache control for reverse proxy

94noni:
hello 👋🏻 

this is kind of a following PR of https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/638

---

we are in the process of leveraging fastly, and fastly waits same value as the `Cache-Control` one, thus we need some array values not one value
[see doc](https://docs.fastly.com/en/guides/controlling-caching#setting-different-ttls-for-fastly-cache-and-web-browsers) [and doc](https://www.fastly.com/documentation/reference/http/http-headers/Surrogate-Control/)

actually, its only stated to have an `integer` in [the doc](https://foshttpcachebundle.readthedocs.io/en/stable/reference/configuration/headers.html#reverse-proxy-ttl) for `reverse_proxy_ttl` which they translates to the proper header name for fastly

so as a hack for now, we arrive to do such things as:

`'reverse_proxy_ttl' => 'max-age=' . 2592000 . ' , public',` 

obviously, I do prefer to have a proper new config map (related to the normal `cache_control`):

```
'reverse_proxy_cache_control' => ['max_age' => 2592000, 'public' => true,],
```

for now no test or whatsoever, I just want to discuss the topic, so opening as an rfc

thx !

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -337,6 +337,15 @@ final class Configuration implements ConfigurationInterface
                         ->defaultNull()
                         ->info('Specify a custom time to live in seconds for your caching proxy. This value is sent in the custom header configured in cache_control.ttl_header.')
                     ->end()
+                    ->arrayNode('reverse_proxy_cache_control')
+                        ->info('Add the specified cache control directives for reverse proxy.')
+                        ->children()
+                            ->scalarNode('max_age')->end()
+                            ->scalarNode('s_maxage')->end()
+                            ->booleanNode('private')->end()
+                            ->booleanNode('public')->end()
+                        ->end()
+                    ->end()
                     ->arrayNode('vary')
                         ->beforeNormalization()->ifString()->then(function ($v) {
                             return preg_split('/\s*,\s*/', $v);
--- a/Resources/doc/reference/configuration/headers.rst
+++ b/Resources/doc/reference/configuration/headers.rst
@@ -347,6 +347,39 @@ section:
 
 This example adds the header ``X-Reverse-Proxy-TTL: 3600`` to your responses.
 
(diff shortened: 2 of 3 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -333,10 +333,18 @@ private function addCacheControlSection(ArrayNodeDefinition $rootNode): void
                         ->end()
                         ->info('Set a default last modified timestamp if none is set yet. Value must be parseable by DateTime')
                     ->end()
-                    ->scalarNode('reverse_proxy_ttl')
-                        ->defaultNull()
-                        ->info('Specify a custom time to live in seconds for your caching proxy. This value is sent in the custom header configured in cache_control.ttl_header.')
-                    ->end()
+                    ->arrayNode('reverse_proxy_ttl')
```
94noni, line 336: > [!NOTE]
> very much trying here, I am not fluent on bundle config at all

94noni:
@dbu friendly ping as reviewer/merger of my previous PR :) 
thank you

dbu:
hm, i see. this starts to go into the territory of https://github.com/FriendsOfSymfony/FOSHttpCache/issues/455 which led to https://github.com/symfony/symfony/issues/47288, which i tried to do in https://github.com/symfony/symfony/pull/59219

i pinged the symfony maintainers on the pull request to see if we can move that forward.

94noni:
I was not really aware of all those, as its quite récent we want to add such cache layer on our app
Will read it carefully thx

94noni:
Lets close here as wont merge for now
Will see if topic advance on symfony side
On my side we handled with internal listener

Thx

PR #645, opened 2025-06-06
Add support for single action controllers, with __invoke method

pawel-kowalczyk-alpy:
PR sponsored by [AlpinResorts](https://www.alpinresorts.com)

### DIFF
--- a/src/EventListener/AttributesListener.php
+++ b/src/EventListener/AttributesListener.php
@@ -27,12 +27,16 @@ final class AttributesListener implements EventSubscriberInterface
         $request = $event->getRequest();
         $controller = $this->controllerResolver->getController($request);
 
-        if (!is_array($controller) || 2 !== count($controller)) {
+        if (true === \is_object($controller)) {
+            $class = new \ReflectionClass($controller);
+            $method = $class->getMethod('__invoke');
+        } elseif (true === \is_array($controller) && 2 === \count($controller)) {
+            $class = new \ReflectionClass($controller[0]);
+            $method = $class->getMethod($controller[1]);
+        } else {
             return;
         }
 
-        $class = new \ReflectionClass($controller[0]);
-        $method = $class->getMethod($controller[1]);
         $attributes = [];
         $addAttributes = static function ($instance) use (&$attributes) {
             if ($key = match (get_class($instance)) {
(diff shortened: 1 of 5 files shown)


pawel-kowalczyk-alpy:
Looks like PHPStan jobs are broken, not related to my changes

pawel-kowalczyk-alpy:
Friendly ping to @dbu

src/EventListener/AttributesListener.php
```diff
@@ -27,12 +27,23 @@ public function onKernelRequest(RequestEvent $event): void
         $request = $event->getRequest();
         $controller = $this->controllerResolver->getController($request);
 
-        if (!is_array($controller) || 2 !== count($controller)) {
+        if (false === $controller) {
```
dbu: this seems a bit dangerous to me. what if the `$controller` is something weird, like a different kind of array, or some non-array non-object but not false?

should we not add the is_object condition here? that should also avoid the uncertainty below whether method and class are null or set.

```suggestion
        if (!is_object($controller) && (!is_array($controller) || 2 !== count($controller))) {
```

or we could do an if object, elseif array, else return.

dbu (commented):
thanks for looking into this, we should indeed support __invoke controllers.

i will have a look at the phpstan failure.

src/EventListener/AttributesListener.php
pawel-kowalczyk-alpy: Return type for `\Symfony\Component\HttpKernel\Controller\ControllerResolverInterface::getController` is `callable|false`, so return values are limited, but you are right, there can be `string` for example.

I've changed a logic, now it will work only for `object` and `array` with two elements

dbu (approved):
great, thanks a lot! i like to be very defensive with these things, you never know what explodes randomly for people...

pawel-kowalczyk-alpy:
Nice! Any chance for a release soon? Thanks anyway

dbu:
yes, "soon" :-) i hope that we can wrap up #646 and include that in the release. if that does not happen this week, i will release what we have.

pawel-kowalczyk-alpy:
Sure, good luck :)

dbu:
we have been quick :-D here you go: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/3.2.0

pawel-kowalczyk-alpy:
Thank you

PR #646, opened 2025-06-16
Add support for configurable HTTP request factory for proxy clients

jdreesen:
It would be great if you could configure the third parameter of the `\FOS\HttpCache\ProxyClient\HttpProxyClient` without having to overwrite the entire service.

### DIFF
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -405,7 +405,7 @@ configure a service for the ``HttpClient`` and specify that in the
 .. _custom HTTP request factory:
 
 Custom HTTP Request Factory
-------------------
+---------------------------
 
 The proxy client uses an implementation of ``Http\Message\RequestFactory`` to create HTTP requests.
 If you need to customize the request creation, you can configure your custom service and
--- a/Resources/doc/reference/configuration/proxy-client.rst
+++ b/Resources/doc/reference/configuration/proxy-client.rst
@@ -402,6 +402,15 @@ example to send a basic authentication header with each request, you can
 configure a service for the ``HttpClient`` and specify that in the
 ``http_client`` option of any of the cache proxy clients.
 
+.. _custom HTTP request factory:
+
+Custom HTTP Request Factory
+------------------
+
+The proxy client uses an implementation of ``Http\Message\RequestFactory`` to create HTTP requests.
+If you need to customize the request creation, you can configure your custom service and
+specify that in the ``request_factory`` option of any of the cache proxy clients.
+
 Caching Proxy Configuration
 ---------------------------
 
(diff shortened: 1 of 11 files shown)


src/DependencyInjection/FOSHttpCacheExtension.php
```diff
@@ -430,6 +430,12 @@ private function loadVarnish(ContainerBuilder $container, XmlFileLoader $loader,
         $container->setParameter('fos_http_cache.proxy_client.varnish.options', $options);
 
         $loader->load('varnish.xml');
+
+        $requestFactory = isset($config['http']['request_factory'])
+            ? new Reference($config['http']['request_factory'])
+            : null;
+        $container->getDefinition('fos_http_cache.proxy_client.varnish')
+            ->replaceArgument(2, $requestFactory);
```
dbu, line 438: i would inline this to make things more compact:

```suggestion
            ->replaceArgument(2, $config['http']['request_factory'] ?? null);
```

(also in the other places below)

dbu (commented):
yeah, seems a reasonable idea to have this.

can you please add this to the documentation as well so the option is explained? and a note in the changelog would be nice.

and if you have time to add a test that ensures the factory is actually configured when specified in the config, that would be nice too.

src/DependencyInjection/FOSHttpCacheExtension.php
jdreesen: I'd love to do this, but it has to be wrapped in `new Reference(...)` when it's set, and I'm afraid that won't work with the null coalescing operator.

dbu: you are correct, sorry.

jdreesen:
> can you please add this to the documentation as well so the option is explained? and a note in the changelog would be nice

Will do that.

> (also for the tweak with the ProxyClient alias :-) ) the version will be 3.2.0

Please note that this PR targets `2.x` (because we're currently stuck with Symfony 5 :disappointed: for the other PR I forgot to select `2.x` as the target :facepalm:, but this one is more important for us).

> and if you have time to add a test that ensures the factory is actually configured when specified in the config, that would be nice too.

I'll look into it.

dbu:
oh, i did not realize this goes to 2.x

alright, i can do a new minor of 2.x for this to help you out. i will wait until you added the doc, changelog and if you manage to add a test.
for the proxyclient i added the changelog entry meanwhile.

Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -402,6 +402,15 @@ example to send a basic authentication header with each request, you can
 configure a service for the ``HttpClient`` and specify that in the
 ``http_client`` option of any of the cache proxy clients.
 
+.. _custom HTTP request factory:
+
+Custom HTTP Request Factory
+------------------
```
dbu:
```suggestion
---------------------------
```

jdreesen: Oops, thx!

dbu:
there you go https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/2.18.0

thanks for the contributions!

jdreesen:
Wow, that was fast! Thank you very much :green_heart:

PR #647, opened 2025-06-16
Set `FOS\HttpCache\ProxyClient\ProxyClient` class alias to default proxy client

jdreesen:
When only `cache_manager.custom_proxy_client` but no `proxy_client.*` is configured.


dbu:
makes sense, thanks

PR #649, opened 2025-06-17
merge 2.x to 3.x

dbu:


Resources/doc/reference/configuration/proxy-client.rst
```diff
@@ -400,6 +400,15 @@ example to send a basic authentication header with each request, you can
 configure a service for the ``HttpClient`` and specify that in the
 ``http_client`` option of any of the cache proxy clients.
 
+.. _custom HTTP request factory:
+
+Custom HTTP Request Factory
+---------------------------
+
+The proxy client uses an implementation of ``Http\Message\RequestFactory`` to create HTTP requests.
```
jdreesen: I think this is different for `FOSHttpCache` `v3` (it changed to `Psr\Http\Message\RequestFactoryInterface`) but `v3` of the bundle supports `v2` and `v3` of the lib. How do we properly document that?

jdreesen: See `v2`: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/2.x/src/ProxyClient/HttpProxyClient.php#L59
vs `v3`: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/src/ProxyClient/HttpProxyClient.php#L51

src/DependencyInjection/Configuration.php
```diff
@@ -624,6 +624,10 @@ private function getHttpDispatcherNode(): ArrayNodeDefinition
                     ->defaultNull()
                     ->info('Httplug async client service name to use for sending the requests.')
                 ->end()
+                ->scalarNode('request_factory')
+                    ->defaultNull()
+                    ->info('Service name of factory for PSR-7 messages.')
```
jdreesen: I'm a bit confused right now... is it `PSR-7` or `PSR-17`?

The parameter description in `v3` says `PSR-7` https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/src/ProxyClient/HttpProxyClient.php#L43
But the Discovery class is named `PSR-17`: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/src/ProxyClient/HttpProxyClient.php#L55

src/Resources/config/varnish.xml
```diff
@@ -10,6 +10,7 @@
                  public="true">
             <argument type="service" id="fos_http_cache.proxy_client.varnish.http_dispatcher"/>
             <argument>%fos_http_cache.proxy_client.varnish.options%</argument>
+            <argument /> <!-- request factory -->
```
jdreesen, line 13: We _could_ use a new Symfony 5.1 feature here: https://symfony.com/blog/new-in-symfony-5-1-abstract-service-arguments

I don't know what your strategy with this this.

src/DependencyInjection/Configuration.php
dbu: oh, good catch! i think that has historical reasons: when we started with this, there was no PSR-17 yet and we used some custom factory if i remember correctly. meanwhile, we completely switched to PSR-17, which is the factory for the PSR-7 http messages.

src/Resources/config/varnish.xml
dbu: ah right, that is a thing now. but as `null` is a valid argument here, i would keep this as is.

src/DependencyInjection/Configuration.php
dbu: i adjusted the documentation in this MR

Resources/doc/reference/configuration/proxy-client.rst
dbu: good point. i actually don't see value in keeping to support FOSHttpCache 2.x with the 3.x bundle. both have the same minimum PHP version, so there is no good argument to use different major versions.

i will adjust accordingly.

dbu: ah, and we also added a psr-17 stream factory optional argument. for consistency we should support that as well.

jdreesen: > ah, and we also added a psr-17 stream factory optional argument. for consistency we should support that as well.

Should I create a PR for that?

dbu: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/595 for improving the phpdoc vor FOSHttpCache 3.x

dbu: thanks. i already committed it in this branch :-D

PR #651, opened 2025-12-03
allow installation with symfony 8 and test with php 8.5

dbu:
* we need to rewrite all service configuration from xml to php (use [this tool](https://github.com/GromNaN/symfony-config-xml-to-php) to automate that mostly)

PR #652, opened 2025-12-22
Use converter to convert xml to php

alexander-schranz:
Did let the converter run to the XML files while do some PRs on the converter Pull Request. I did not yet implement any loading of the new files, leak of time. But maybe we can merge this already into the sf8 Branch so somebody can continue there. https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/651


dbu:
thank you. i wonder if we need to keep the xml files or if we can switch to use the php loader with all versions of symfony

PR #654, opened 2026-01-12
Load PHP files in DI, remove XML config

emodric:
This switches DI to actually load PHP files instead of XML and removes the old XML config.

Note: I did not test this on Symfony 6.4, only 7.4 and 8.0, but 6.4 does have `PhpFileLoader`, so presumably it should work fine.


dbu:
thanks a lot! 
can you please adjust https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/389b6d84d6fc754279eb52cbdd57f1952907dc0e/tests/Unit/DependencyInjection/ConfigurationTest.php#L44 to only load the empty.xml (and the other places we do that) if we are on symfony < 8?

not sure about the security advisory errors. i think we should simply disable the composer audit function - it is up to the applications to respect or ignore them. and symfony 6.4 is still supported, so i hope a fixed version releases soon.

the console application add is a different issue, would need to check what the new way of adding commends is. we can do that in a separate MR.

emodric:
> to only load the empty.xml (and the other places we do that) if we are on symfony < 8?

~Is this really needed? I believe only DI component removed support for XML format. Config component still supports XML as per Symfony docs.~

Ah, ok, I see it now, Config component depends on DI.

> not sure about the security advisory errors. i think we should simply disable the composer audit function - it is up to the applications to respect or ignore them. and symfony 6.4 is still supported, so i hope a fixed version releases soon.

As for this, sure I can disable security advisories. However, I tried installing Symfony 6.4 locally with PHP 8.2 and it works fine, no advisory errors given, so I'm not really sure why it fails here.

emodric:
> the console application add is a different issue, would need to check what the new way of adding commends is. we can do that in a separate MR.

It turns out it's just `addCommand` instead of `add`, so I fixed that too.

dbu (approved):
great, thanks a lot!

PR #657, opened 2026-05-28
Bugfix: automatically detect url type generation for varnish

mvhirsch:
Dear @dbu,

I now stumbled upon a bug where `generate_url_type` must be set if you're using Varnish and `proxy_client.varnish.http.base_url`  is set. This is because the extension only takes `proxy_client.[client].base_url` into account, while Varnish client uses `proxy_client.varnish.http.base_url` (lacking "http" in array selector).

This PR fixes the automatic absolute/relative path generation if you use a combination of Varnish + custom base URL.


dbu (approved):
thank you for this bugfix

